PR #21, opened 2026-04-23
Add first-boot setup UI

bokiko:
## Summary
- add a dedicated `/setup` page for the first-boot flow with setup token, admin username, password, and terminal PIN entry
- auto-check `GET /api/setup/status` so the dashboard redirects to `/setup` before login when initialization is still pending
- auto-login and hand off into the normal dashboard flow after a successful setup submission
- allow unauthenticated access to `/setup` while preserving the existing auth guard for the rest of the app
- update README first-boot steps to match the new product flow instead of raw API calls

## Checks
- `go test -count=1 ./...` in `hub`
- `go test -count=1 ./...` in `agent`
- `pnpm lint` in `dashboard`
- `pnpm build` in `dashboard`

## Notes
- if setup is already complete, `/setup` redirects back into the normal login or dashboard path
- the backend setup contract is unchanged; this PR turns it into a real first-boot product flow

### DIFF
--- a/HANDOFF.md
+++ b/HANDOFF.md
@@ -23,17 +23,18 @@
     - [x] GitHub Actions CI for hub tests, agent tests, dashboard lint/build
 - In progress:
   - [ ] PR `#21` / branch `feat/setup-ui`: frontend `/setup` page for first-boot flow
-    - current branch head: `1692c2d`
+    - current branch head: `3fcb836`
     - latest fixes included:
       - blocking error state when setup status cannot be fetched
       - explicit message when setup succeeds but auto-login fails
       - login page skips setup probe when already authenticated
       - setup retry button retriggers the setup-status fetch correctly
+      - RBAC/action-scope foundation (`admin` / `operator` / `viewer`) with server-side route checks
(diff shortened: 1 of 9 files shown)


bokiko:
Good review. I agree with the two error-path issues and the optional login short-circuit.

I'll patch before merge:

- `GET /api/setup/status` failure will become a blocking error/retry state, not fall through to the setup form.
- After successful `POST /api/setup`, if auto-login fails, the page will show an explicit error/banner and keep the manual login path visible.

I'll also include the small login-page guard so the setup-status probe doesn't run when already authenticated.

I'm fine leaving the setup token field as-is for now since it's single-use, but masking it later is reasonable.

bokiko:
Fetched `feat/rbac-route-audit` on Linux and tested it. `go test -count=1 ./...` in `hub` passes. The branch only changes `hub/main.go`, `hub/main_test.go`, and `hub/rbac.go`; no dashboard files changed. I couldn't rerun dashboard in the fresh worktree because it had no `node_modules`, but dashboard was already green on base `5611c80`. Looks good to land `3fcb836`.

PR #94, opened 2026-05-13
Fix machine delete credential cleanup

bokiko:
## Summary
- Delete durable `agent_credentials` rows when deleting a machine
- Add a regression test proving machine deletion invalidates the stored agent credential

## Verification
- `go test -count=1 -run 'TestDeleteMachineRemovesAgentCredential|TestOperatorCanUpdateTagsButCannotDeleteMachine' ./...`
- `go test -count=1 ./...`

git diff --check passed locally.

### DIFF
--- a/hub/main.go
+++ b/hub/main.go
@@ -110,7 +110,6 @@ type CommandResponse struct {
 	Error   string `json:"error"`
 }
 
-
 // TerminalSession tracks an active terminal relay session.
 var (
 	db       *sql.DB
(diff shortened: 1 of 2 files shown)


bokiko:
LGTM. Adds agent_credentials to the transactional delete cascade — verified no FK ON DELETE CASCADE on that table so explicit delete is required. Test exercises the real HTTP handler.

Suggested follow-ups (separate PRs, non-blocking):
1. Schema migration to add ON DELETE CASCADE to agent_credentials so future credential-bearing tables can't be forgotten.
2. A regression test that the deleted secret is rejected by the agent enrollment endpoint, not just gone from the table.
3. Audit handleDeleteAPIMachine for the same class of leftover-credential bug on api_machines.auth_config.

Issue #60, opened 2026-05-03
tests: race-detector finds db-global access after test teardown


bokiko:
## Current finding

`go test -race -count=1 ./...` in `hub/` consistently reports a race between test setup replacing package-level server state and asynchronous agent WebSocket goroutines from an earlier test continuing to use that state.

The write is `setupTestServer` assigning the package-global database handle. Observed readers vary by scheduler and platform and have included:

- `storeMetrics`
- `markOffline` through `unregisterAgentConnection`
- `lookupMetricsOS` through asynchronous version announcement

The changing reader with the same write site proves this is not a single-call-site bug and should not be hidden with a longer sleep or a cleanup added to only one test.

## Impact

- **Production:** no demonstrated production race; the database handle is assigned once during startup.
- **Tests and CI:** the full race suite is permanently red, making the advisory race job easy to ignore and preventing it from becoming a trustworthy merge gate.
- **Architecture:** asynchronous handlers can outlive the test that installed the package-global state they reference, preventing isolated server instances and safe parallel testing.

## Correct direction

Introduce server-instance ownership for the database handle and the asynchronous WebSocket paths that use it. Tests should construct a server bound to their own database instance rather than reassigning package-global state.

The implementation must also provide explicit shutdown or lifecycle control so test-owned WebSocket goroutines cannot outlive their server instance.

Do not solve this with:

- longer sleeps;
- a fix limited to one observed reader;
- an `atomic.Pointer` around a process-wide database global;
- leaving the race job permanently non-blocking.

## Acceptance criteria

- [ ] Server and WebSocket handlers use the database instance belonging to the registered server.
- [ ] Asynchronous metrics, disconnect, OS lookup, and version-announcement paths retain the originating server instance.
- [ ] Tests construct isolated server instances without replacing a database handle still reachable by earlier goroutines.
- [ ] Test shutdown waits for or cancels server-owned asynchronous work.
- [ ] `cd hub && go test -race -count=1 ./...` completes without a race report.
- [ ] Normal hub tests and builds remain clean.
- [ ] The `hub-race` CI job has `continue-on-error: true` removed.
- [ ] The check is renamed from `Hub Tests (race, non-blocking)` to `Hub Tests (race)`.
- [ ] After the renamed check exists and passes on the fix PR, add `Hub Tests (race)` to the required checks for `main` as a separate repository-settings gate.

## Current status

[...]

PR #134, opened 2026-07-31
ci: add race, vet, windows, and insecure-tag coverage

bokiko:
## Why

CI cannot currently observe several bug classes this repo already has open. It runs a bare `go test -count=1 ./...` on two Linux jobs and lint+build on the dashboard — no race detector, no `go vet`, no Windows target, and no build with the `insecure` tag.

This lands the net before the fixes, so the P0 work in flight has something to fall into.

## What changed

| Job | Change |
|---|---|
| `hub` | added `go vet ./...` |
| `hub-race` | **new** — `go test -race -count=1 ./...`, `continue-on-error: true` |
| `agent` | added `go vet ./...` and `go vet -tags insecure ./...` |
| `agent-windows` | **new** — `windows-latest`, vet + `go test ./...` in `agent/` |
| `dashboard` | untouched |

### `hub-race` is non-blocking on purpose

It fails today, and that failure is [#60](https://github.com/bokiko/bloxos/issues/60), not a flake. Reproduced locally at `78608e9`:

```
--- FAIL: TestAgentWSSecretQueryFallback (1.31s)
WARNING: DATA RACE
Write at 0x0001022b41c0 by goroutine 2156:
  hub.setupTestServer()      hub/main_test.go:131
  hub.TestAgentWSSecretQueryFallback()  hub/ws_credential_headers_test.go:113

Previous read at 0x0001022b41c0 by goroutine 2152:
  hub.storeMetrics()         hub/main.go:840
  hub.handleAgentWS()        hub/agentws.go:684
```

`hub/main.go` keeps the `*sql.DB` handle and ten other pieces of state in package globals. `setupTestServer` reassigns them per test while the previous test's leaked agent-WS goroutine is still writing metrics through them. That is an architecture fix (a server struct), not a test fix — so making this a required check today would block every unrelated PR.

**`continue-on-error: true` comes off the moment #60 lands.** There's a comment in the workflow saying exactly that.

### The `insecure` tag built nowhere

`agent/tls_insecure.go` and `agent/tls_insecure_test.go` are both `//go:build insecure`. A bare `go test ./...` builds neither. That file and its test have never been compiled by CI — including in #129, the Phase 2E change that introduced them. One `go vet -tags insecure` step closes it.

### Windows

~1,600 lines of `agent/*_windows.go` were never compiled here. A `windows-latest` runner costs about the same as a `GOOS=windows` cross-build and additionally runs the tests. Worth being explicit: a compile gate would **not** have caught the known Windows updater defects (service never restarts after update, `applyPendingUpdate` skips SHA re-verification, the helper batch deletes before it moves, `restart_service` doesn't wait for STOPPED). Those need real tests and one manual run on hardware. This is the floor, not the ceiling.

### Not included: `GOOS=darwin`

`GOOS=darwin go build ./...` fails in `agent/` with 10+ undefined symbols — there are no `*_darwin.go` files at all. macOS agent support is a product decision, not a CI line, so no darwin gate here.

## Verification at `78608e9`

- `go vet ./...` — clean in `hub/`, and in `agent/` for both `GOOS=linux` and `GOOS=windows`
- `go vet -tags insecure ./...` — clean in `agent/`
- `go test -race -count=1 ./...` in `hub/` — 1 data race, 1 test failure, as quoted above
- workflow parses as YAML; jobs resolve to `hub, hub-race, agent, agent-windows, dashboard`

`agent/` cannot be built or tested natively on macOS, so no local test-pass claim is made for it — that's what the new Linux/Windows jobs are for.

Co-authored-by: bokiko <bokiko@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,36 @@ jobs:
           go-version: "1.25.9"
           cache-dependency-path: hub/go.sum
 
+      - name: Vet hub
+        run: go vet ./...
+
       - name: Run hub tests
         run: go test -count=1 ./...
+
+  # Reports only; does not block merges yet. The hub keeps its DB handle and ten
+  # other pieces of state in package globals (hub/main.go), so main_test.go
+  # reassigns them between tests while leaked agent-WS goroutines still write
+  # through them -- that is issue #60, and it is an architecture fix, not a test
+  # fix. Flip continue-on-error off once #60 lands so this becomes a real gate.
+  hub-race:
+    name: Hub Tests (race, non-blocking)
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    defaults:
+      run:
+        working-directory: hub
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.25.9"
+          cache-dependency-path: hub/go.sum
+
+      - name: Run hub tests with the race detector
+        run: go test -race -count=1 ./...
 
   agent:
     name: Agent Tests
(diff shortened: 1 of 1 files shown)


bokiko:
## CI result — the race job behaves as designed

| Check | Result |
|---|---|
| Hub Tests | pass |
| Agent Tests | pass (includes the new `-tags insecure` test step) |
| Agent Tests (windows) | pass |
| Dashboard Lint + Build | pass |
| **Hub Tests (race, non-blocking)** | **fail — this is #60** |

Workflow run conclusion is `success`, so `continue-on-error` works as intended.

**Two things to be precise about, because "non-blocking" is doing some work in that job name:**

1. The check still renders **FAILURE** in this PR's status rollup — non-blocking means non-gating, not green. Confirmed against branch protection: required checks are `Hub Tests`, `Agent Tests`, `Dashboard Lint + Build`. The race job is not among them, so it cannot gate a merge.
2. **`Agent Tests (windows)` is also not a required check.** This PR adds the job but cannot add it to the required list — that's a repo setting. Until someone does, a Windows-only regression shows red and merges anyway, which is most of the value of the job. Recommend adding `Agent Tests (windows)` to required checks when this lands, and leaving the race job off until #60 is fixed.

The CI race stack differs from the local one — same write site, different reader (`markOffline` via the disconnect path rather than `storeMetrics`). Posted in full on #60, since it strengthens the case that #60 is an architecture problem rather than one guardable call site.

Issue #130, opened 2026-07-15
Post-hardening runtime shakedown pending


bokiko:
# Post-hardening runtime shakedown

**Latest:** follow-up `50d6a12` is pushed and deployed, hub/dashboard only. Both containers are healthy, all three machines are online with advancing non-degraded history ACKs, and served agent binaries are unchanged. New-head CI is green except the race job still running. Squash auto-merge is queued but requires GitHub maintainer approval; no bypass. Issue remains open until PR182 merges. This supersedes the deployment/auto-merge status below.

Updated 2026-09-07. Runtime findings have been diagnosed and the follow-up fixes are deployed to the evaluation stack from PR #182 (`7cf37c9`). Keep this issue open until that PR lands. Review follow-up `50d6a12` is now pushed: cancellation guards for single/bulk commands and non-2xx feedback, with passing targeted race regressions and 22 dashboard tests/lint/TypeScript/build. Fresh CI/re-review and candidate build are pending; auto-merge remains disabled during those checks. Live remains on the verified candidate. No branch-protection bypass was used.

## Verified runtime and fixes

- [x] Original staged release-3 rollout completed: ai-03 through trusted admin provisioning; bench/dualasus through the existing signed self-update path. All three retain exact binary identities, pinned keys and release floors. Hub rollback does not undo an agent floor.
- [x] Fresh protected data backup before this follow-up deployment; archive checksum and copied database integrity/schema checked. Earlier full data/CA/config/image backups and rollback tags retained. Hub/dashboard recreated with image-only overrides; Caddy, CA, signing keys, credentials, previews and existing agent binaries unchanged.
- [x] New hub/dashboard are healthy. Trusted-CA HTTPS health and login succeed; unauthenticated machine/version APIs still return 401. All three machines are online with acknowledged power-history streams. Independent read-only queries of the live WAL database show advancing cursors and no degraded streams.
- [x] **SSE diagnosis:** a fresh stream lasted approximately 241 seconds, matching its normal four-minute token refresh. The apparent warning was an invisible `opacity: 0; height: 0` exit-animation node lingering in an automation-hidden tab, not a failing stream. No speculative SSE rewrite was applied.
[...]

PR #135, opened 2026-07-31
fix(update): require authenticated transport and a pinned signing key for agent self-update

bokiko:
Closes the unauthenticated-root-RCE in the agent self-update path, and moves the systemd start-limit keys to their documented location.

Base: `main` @ `8a38f1b` (rebased onto #134 so the Windows and race jobs run against this diff).

## 1. The hole

`agent/updater.go` derived the download scheme from the hub URL: **only `wss` yielded `https`, everything else yielded plain `http`.** On an HTTP-served hub the entire chain was unauthenticated:

- `hub/agentws.go` maps `https→wss` only, so an http hub hands every agent `BLOXOS_HUB=ws://…`
- `/download/agent` sits in the **"Public endpoints (no auth)"** block (`hub/main.go:256`)
- the SHA authorising the swap arrives over that same plaintext WebSocket, and `updater.go:129` only checked the download against it — a corruption check, not an authenticity check. No signature, no pinned key.
- the hub-generated unit runs the agent as `User=root` (`hub/agentws.go:233`)

Anyone on the network path between an agent and the hub could serve their own binary plus a matching SHA and take root on every machine in the fleet, bypassing the command allowlist the rest of the agent is built around.

## 2. Two gates, both cleared before a single byte is downloaded

**Transport** — `agentDownloadURL` (`agent/update_transport.go`) refuses `ws://` and `http://` outright. Loopback is exempt: the hub is on the same machine, so there is no network path to sit on, and `ws://localhost:4000` is the shipped dev default.

**Authenticity** — the hub signs `bloxos-agent-update:v1:<os>:<sha256>` with ed25519 and the installer pins the public half on the agent:

| | pinned at |
|---|---|
| Linux | `/etc/bloxos/agent-update.pub` (install.sh) |
| Windows | `<InstallDir>\agent-update.pub`, beside the .exe (install.ps1) |

The agent refuses any announcement it cannot verify against that key — including one carrying no signature at all. The OS is inside the signed message so a signature issued for the Linux binary cannot be replayed as the Windows announcement.

**Fail-closed.** No pinned key means no self-update, recoverable by re-running the hub installer. The hub likewise declines to *announce* a version it cannot sign, rather than announcing one every agent will reject and then tripping the rollout circuit breaker on the reconnect it waits for.

## 3. Where the signing key lives

Priority order per announcement:

1. **Detached `<binary>.sig`** next to the agent binary — produced offline at build time, so the private key never touches the hub. Set `BLOXOS_UPDATE_PUBKEY` and the hub validates the `.sig` before announcing, so a stale one is caught hub-side instead of by the whole fleet.
2. **Hub-held key** — `BLOXOS_UPDATE_SIGNING_KEY` (path to a base64 key or seed), else generated and persisted at `~/.bloxos/update-signing.key` on first boot, mirroring `loadOrGenerateJWTSecret`.

**Threat model, stated plainly:** mode 2 does not defend against a compromised hub — nothing signed on the hub can. What both modes buy is that the public unauthenticated `/download/agent` route can no longer be substituted by anything that is not the hub, which together with the transport gate closes the network-path attack. Mode 1 additionally survives hub compromise, and is why the detached-signature path exists.

**Downgrade:** there is still no version monotonicity, so a *hub* could announce an older signed release. A network attacker can no longer replay one, because the announcement now only arrives over TLS from the hub. Monotonicity needs persistent per-agent state and is deliberately left out of this PR.

### When the hub announces an update at all

Signature verification lives in the agent binary, so it cannot be retrofitted
into one already deployed. A `78608e9` agent has only `type`/`sha256`/`version`
in its `AgentVersionMessage` — `encoding/json` silently discards the signature —
and it goes straight to `performUpdate`. That one migration hop is
unauthenticatable by construction. Credit to Codex for catching it.

What the hub *can* control is whether it announces at all. Two rules, both
fail-closed:

**1. Never announce to an agent the hub has heard nothing from.**
`announceVersionToAgent` runs at WS-upgrade time, before
`agent_running_version` arrives, so on a first connect the hub knows neither
the agent's capability nor its transport. Every agent sends that frame on
connect and `recordAgentRunningVersion` re-triggers the announce, so waiting
costs one round trip and removes the guess entirely.

**2. Never announce an update the recipient will refuse.** The rule was
already in this function for the matching-SHA case — *never arm the reconnect
timer for an update you can predict won't produce a reconnect* — and the
transport case is a second instance of it. Announcing to a refusing agent arms
a 90s reconnect expectation that expires into a rollout failure; **two of those
trip the process-wide circuit breaker and pause updates for the whole fleet**,
with `rolloutPauseReason` blaming agent health for a refusal the hub provoked.

| Agent | Signal used | Announce |
|---|---|---|
| Never reported | — | **withheld** until it reports |
| `update_protocol >= 1`, reports `update_transport_ok: true` | the agent's own answer | sent, signed |
| `update_protocol >= 1`, reports `update_transport_ok: false` | the agent's own answer | **withheld** |
| Pre-signature, `PUBLIC_URL` https/wss or loopback | `PUBLIC_URL` | migration announce sent |
| Pre-signature, `PUBLIC_URL` remote plaintext or unset | `PUBLIC_URL` | **withheld** — re-run the installer |

**The capable agent is asked rather than guessed at.** `PUBLIC_URL` is the hub
inferring the agent's transport, and in a mixed deployment — `PUBLIC_URL`
https, some agents hand-configured with `ws://` — the inference is wrong for
precisely the machines that will refuse, and nothing else distinguishes them.
The agent already computes the answer in `agentDownloadURL` from the
`BLOXOS_HUB` it is actually connected to, so it now reports
`update_transport_ok` alongside `update_protocol`.

Pre-signature agents have no transport gate and nothing to ask, so they still
fall back to `PUBLIC_URL` — and there the reason is security rather than
futility: that unverifiable hop must only happen where an off-host attacker
cannot reach it. `PUBLIC_URL` is the right proxy for them because it is what
`buildInstallCommand` bakes into `BLOXOS_HUB`.

### Every withheld update says why, in the API

Each rule above is a way for fleet updates to stop dead, and on the dashboard
that is indistinguishable from a rollout in progress: `update_pending` true on
every agent, forever. `GET /api/versions` now carries:

- `signing_enabled` / `signing_disabled_reason` — from the state
  `initUpdateSigning` computes, so an unreadable key is diagnosable without
  reading hub stdout
- `update_blocked_reason` per machine — why *this* agent is not being
  announced to

`announceDecision` is the single source of truth for both the API and the
wire, so what an operator is told and what the hub actually does cannot drift.

**`announceDecision` takes no locks.** It receives the caller's already-read
`agentVersionInfo` (nil when the hub has never heard from that machine). The
first version looked the machine up itself, so `handleListVersions` — which
iterates `agentRunningVersions` under `RLock` — re-entered the same `RWMutex`.
`sync.RWMutex` does not permit recursive read locking: a writer queueing
between the outer and inner acquisition blocks the reader, which never
releases, which blocks the writer. `recordAgentRunningVersion` runs on the
agent WebSocket read path, so the whole rollout subsystem would stall
process-wide until a hub restart — triggered by a dashboard poll racing an
agent reconnect. Both routine. Found by Codex.

`lookupVersionHostname` (a `db.QueryRow` per machine) also moved outside that
lock. Pre-existing, but holding a lock the agent-WS path writes to across I/O
was tolerable when the loop was three field assignments and is not now.

### Fleet-wide behaviour change

Agents already in the field have no pinned key, so **self-update stops for them
until each host re-runs the installer.** Intended fail-closed direction, but it
is fleet-wide and not silent: the agent logs `update: REFUSED — no pinned
update key at …` and the hub logs `no update signature available` /
`not announcing an update it could not verify`.

### Key rotation — known limitation, not fixed here

The agent trusts exactly one local key file and the hub announces exactly one
signature. **Changing or losing the hub key means every pinned agent rejects
every later update until the installer is re-run on each host.** That is
documented behaviour, not a bug that got missed — but it is also not a rotation
protocol, and dual-signing / cross-signed next-key belongs in its own change.
Master has deferred it to the plan as a separate item.

### Availability hardening in the second commit

Three failure modes where a file the hub wrote itself could take down the hub
or strand the fleet:

- **Nothing in `initUpdateSigning` is fatal any more.** A corrupt
  `~/.bloxos/update-signing.key` used to `log.Fatalf` — the hub would not boot.
  It now warns and disables signing. A self-update outage should not be
  escalated into a total one.
- **A read error no longer mints a replacement key.** Only a genuine `ENOENT`
  on the default path counts as first boot. `EACCES`, `EISDIR`, a remounted
  volume, a different `HOME` under systemd — the key probably still exists and
  every agent is still pinned to it. A generated key that cannot be *persisted*
  is now discarded rather than used in memory, because an unpersisted key
  changes on every restart and orphans every agent installed against it.
- **`BLOXOS_UPDATE_PUBKEY_PATH`** is the agent's key-file override. The hub
  reads `BLOXOS_UPDATE_PUBKEY` as a base64 key *value*; on a single-box
  deployment both binaries share one environment, so one name for two meanings
  disabled signing and verification simultaneously.

## 5. systemd `StartLimit*` — hygiene, not a fix

Moved `StartLimitIntervalSec=`/`StartLimitBurst=` from `[Service]` to `[Unit]`.

**I need to correct my own earlier review here.** I reported that the `[Service]` placement made these unknown keys, so the unit never reached `failed`, `OnFailure=bloxos-agent-recover.service` never fired, and a bad update crash-looped forever instead of rolling back. **That is wrong** — systemd keeps `StartLimitInterval=`/`StartLimitBurst=` in `[Service]` as backwards-compat aliases. I measured both shapes on systemd 255 rather than reasoning from the docs:

```
=== probe-svc  (StartLimit* in [Service], the shape at 78608e9) ===
state:      failed/failed
NRestarts:  3
OnFailure fired: YES

=== probe-unit (StartLimit* in [Unit], this PR) ===
state:      failed/failed
NRestarts:  3
OnFailure fired: YES
```

and the effective properties are identical on both:

```
$ systemctl show probe-base.service -p StartLimitIntervalUSec -p StartLimitBurst
StartLimitIntervalUSec=1min
StartLimitBurst=3
$ systemctl show probe-new.service  -p StartLimitIntervalUSec -p StartLimitBurst
StartLimitIntervalUSec=1min
StartLimitBurst=3
```

So the Linux rollback was already working. The change is kept only because `[Unit]` + `StartLimitIntervalSec=` is the documented, non-deprecated spelling.

`systemd-analyze verify` on the generated unit — systemd 255 (255.4-1ubuntu8.16), unit fetched from a running hub over `/install.sh`, with `${HUB}`/`${TOKEN}` substituted and the empty `${AGENT_CA_ENV}` line dropped:

```
$ systemd-analyze verify ./bloxos-agent.service
exit=0
```

(clean on both base and this branch — `verify` does not distinguish them, which is itself part of why the original finding was wrong; the `systemctl show` and OnFailure probes above are the decisive evidence.)

Generated unit on this branch:

```ini
[Unit]
Description=BloxOS Agent
After=network.target
StartLimitIntervalSec=60
StartLimitBurst=3
OnFailure=bloxos-agent-recover.service

[Service]
Type=simple
User=root
Environment="BLOXOS_HUB=wss://hub.example.com:4000/ws/agent"
Environment="BLOXOS_TOKEN=enrollment-token"
ExecStart=/usr/local/bin/bloxos-agent
Restart=always
RestartSec=5

[Install]
WantedBy=multi-user.target
```

## 6. Tests

New, all in the always-compiled (untagged) part of `agent/` so they run on both the Linux and Windows CI targets:

- `TestAuthorizeUpdateRejectsPlaintextTransport` — **a perfectly signed announcement over a `ws://` hub is refused.** This is the headline regression test.
- `TestAuthorizeUpdateRequiresPinnedKey` — no pinned key, no update, even over TLS
- `TestAuthorizeUpdateRejectsForeignSigner` — TLS + a signature from an unpinned key is the substitution attack itself
- `TestAgentDownloadURL*` / `TestIsLoopbackHost` — scheme matrix incl. `ws://127.0.0.1@hub.example.com` and `ws://localhost.attacker.example`
- `TestVerifyUpdateSignature` — wrong OS, wrong SHA, wrong key, missing, garbage, non-hex SHA
- hub side: signature round-trips against the public key the installer pins; both installers embed the key and leave no placeholder; detached `.sig` stale/malformed handling; no signing key ⇒ no announce
- `TestUpdateSigningMessageFormat` exists in **both** packages asserting the same literal, so hub and agent cannot drift apart silently

### Round 2

- `TestSigningKeyNotRegeneratedOnNonENOENTReadError` — the one Master asked for:
  an `EISDIR` at the key path produces **no** new key and does not touch the path
- `TestCorruptSigningKeyDisablesSigningWithoutKillingHub` — reaching the end of
  the test at all is the assertion; the old code called `log.Fatalf` here
- `TestExplicitSigningKeyPathMissingDoesNotGenerate`, `TestSigningKeyGeneratedOnFirstBoot`
  (including that a second load returns the *same* key)
- `TestUpdatePubkeyEnvNamesDoNotCollide` — the hub must not read the agent's var
- `TestLegacyBootstrapAllowed` — scheme/loopback matrix incl.
  `http://localhost.attacker.example`
- `TestAgentIsSignatureCapable` — never-seen and `update_protocol=0` are both
  "not capable"
- End-to-end over a real agent WebSocket:
  `TestLegacyAgentGetsNoAnnounceOverPlaintext`,
  `TestLegacyAgentGetsAnnounceOverTLS` (the control — same helper, same
  `update_protocol=0`, so the plaintext case cannot pass vacuously),
  `TestSignatureCapableAgentGetsAnnounceOverPlaintext`,
  `TestNoSignatureMeansNoAnnounceOnTheWire`
- `TestDownloadRefusalWritesNothing` — closes the gates-before-download ordering
  gap: a refused download creates no file and attempts no connection
- `TestNoSignatureMeansNoAnnounce` was renamed to
  `TestNoSignatureAvailableWithoutSigningKey`. It never called
  `announceVersionToAgent`, so it did not test what its name claimed; the
  on-the-wire claim is now its own test.

### Round 3

- `TestNoAnnounceWhenAgentReportsPlaintextTransport` — **this test previously
  asserted the opposite** and locked in the circuit-breaker bug; Codex caught it
- `TestSignatureCapableAgentWithUsableTransportGetsSignedAnnounce` — the control
- `assertNoReconnectExpectation` on every withholding test: no timer armed, breaker not tripped
- `TestVersionsAPIReportsWhyUpdatesAreWithheld` — end-to-end through `GET /api/versions`
- `TestAgentVersionAnnouncedOnReconnect` now sends `agent_running_version` after
  reconnecting, as a real agent does

### Round 4

- `TestVersionsAPIDoesNotDeadlockWithConcurrentAgentReports` — hammers the
  endpoint against the `agent_running_version` write path under `-race`.
  **Verified non-vacuous:** with the nested `RLock` reintroduced it reports the
  deadlock at its 20s deadline; on this code it passes in ~4s. Its cleanup uses
  `TryLock`, because a blocking cleanup runs *after* `t.Fatal` and swallowed the
  diagnostic entirely on the first control run.

### Verification run

Everything below ran on a Linux host (`Linux 6.17.0-35-generic`, go1.25.9) at commit `54b49ab`, since `agent/` has no darwin platform files and cannot be built or tested on macOS at all:

```
agent:  go build ./...                     OK
        go vet ./...                       OK
        GOOS=windows go build ./...        OK
        GOOS=windows go vet ./...          OK
        go vet -tags insecure ./...        OK
        go test -tags insecure -count=1    ok  github.com/bokiko/bloxos/agent  0.220s
        go test -count=1 ./...             ok  github.com/bokiko/bloxos/agent  0.220s
hub:    go build ./... / go vet ./...      OK
        go test -count=1 ./...             ok  github.com/bokiko/bloxos/hub    27.840s
```

`go test -race` over the new and changed hub tests is clean.

The full hub suite still fails under `-race` on the pre-existing issue #60
(`setupTestServer` reassigning the global `db` while leaked agent-WS goroutines
read it). Runs at `8cf701f` showed two races at the same address and the same
write, with different readers — `lookupMetricsOS` via `announceVersionToAgent`,
and `markOffline` via `unregisterAgentConnection`. Neither reader is code this
PR changes.

**That red check is not a shipping risk, and the distinction matters.** `db` is
assigned in production exactly once — `hub/main.go:156`, at startup — and never
again; that is the only non-test writer in `hub/`. The write side of every #60
race is the *test harness* reassigning a package global between tests. A running
hub has no concurrent writer, so no number of announce goroutines can produce
that race in production.

Worth stating because this PR does plausibly make #60 surface more often **in
CI**: the widened re-announce trigger in `recordAgentRunningVersion` can add one
`go announceVersionToAgent` per connect, and that goroutine reads `db` through
`lookupAgentOS`. More flakiness in the advisory job; no change to the shipped
hub. Credit to Master for drawing the CI/production line — my first framing was
more conservative than the facts warrant.

## What this PR does not do

- **No key rotation protocol.** See §4 — one key, one signature, rotation means
  re-running the installer fleet-wide. Deferred as its own item.
- **No retrofit of verification into already-deployed binaries.** Impossible by
  construction; §4 states what happens to them instead.
- No Windows self-update completion (SCM restart on exit 0, `applyPendingUpdate`
  SHA re-check, move-not-delete in the helper batch). Separate change, and it
  needs one manual run on a real Windows box. Codex flagged this as
  non-blocking and pre-existing; it is untouched here.
- No change to `validTarget`, the unbounded `go handleCommand` fan-out, or the
  hardware-collector timeouts.

---

> [!NOTE]
> **High Risk**
> Changes authentication, signing, and rollout gating for root-running agent self-update; fleet-wide behavior until each host re-runs the installer, with misconfigured signing or transport blocking updates.
> 
> **Overview**
> Closes the **unauthenticated self-update RCE**: agents no longer download over plain HTTP from non-loopback hubs, and they **refuse updates** unless the hub’s `agent_version` announcement carries an **ed25519 signature** over `bloxos-agent-update:v1:<os>:<sha256>` that verifies against an **install-time pinned public key** (`authorizeUpdate` runs before any download).
> 
> **Agent:** New `agentDownloadURL` / `refuseRedirects` gates; Linux and Windows updaters call `authorizeUpdate`, attach `signature`/`sig_alg` to frames, and report `update_protocol`, `update_transport_ok`, and `update_key_pinned` in `agent_running_version`.
> 
> **Hub:** `initUpdateSigning` and `hub/update_signing.go` (hub-held key, `BLOXOS_UPDATE_PUBKEY`, or detached `<binary>.sig`); installers pin the pubkey; **`announceDecision`** withholds announces when signing/transport/key readiness would fail (shared with **`GET /api/versions`** via `update_blocked_reason` and signing status). **`handleListVersions`** no longer holds `agentRunningVersionsMu` across `announceDecision`/DB lookups (deadlock fix).
> 
> **Ops:** Existing agents need **installer re-run** to pin keys; signing key load failures disable signing instead of fataling; systemd `StartLimit*` moved to `[Unit]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89e024f772ffe27b7225c031d47bcf1092649260. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko:
Independent verification pass at `04d353e`. **The authenticity model is sound — I could not find a bypass.** Missing signature, empty signature, missing key file, malformed key, wrong-length ed25519 key, truncated `.sig`, and cross-OS replay all fail closed, and `authorizeUpdate` runs before any bytes are fetched (`agent/updater.go:95` vs `:139`). The transport gate checks `u.Host`, not the raw string, so `ws://127.0.0.1@evil.example.com:4000/` is correctly refused.

Everything below is an **availability** problem — "the fleet can never update again" — not an authenticity bypass.

### 1. `hub/update_signing.go:112-121` — non-ENOENT read error silently rotates the signing key

```go
if data, err := os.ReadFile(keyPath); err == nil {
	...
} else if explicit {
	log.Fatalf(...)
}
_, priv, err := ed25519.GenerateKey(rand.Reader)
```

On the default (non-`explicit`) path this does not distinguish `os.IsNotExist`. Any read failure — EACCES, EISDIR, a container volume remounted, the hub relocated into a systemd unit with a different `HOME` — falls through to generating a brand-new key. The subsequent `os.WriteFile` will usually fail for the same reason, yielding `"(in-memory, not persisted)"`, so **the hub mints a fresh signing key on every restart** while every installed agent still holds the previously pinned one.

Result: fleet-wide self-update stops permanently, with only a per-announce log line, and the sole recovery is re-running the installer on every host. There is no rotation path — the pubkey is emitted only by `/install.sh` and `/install.ps1`.

**Fix:** gate the generate branch on `os.IsNotExist(err)`. This branch is untested.

### 2. `hub/update_signing.go:118` — `log.Fatalf` on a corrupt auto-generated key stops the whole hub

An undecodable `~/.bloxos/update-signing.key` (truncated write, power loss, a stray `echo`) turns into *the hub will not boot* — not just "self-update is unavailable." That is a new whole-hub availability failure mode originating from a file the hub wrote itself.

The comment at `:97-98` says this "mirrors `loadOrGenerateJWTSecret`". It does not: `hub/auth.go:445-449` reads the JWT secret and, on unusable data, silently regenerates and continues. Either match that behaviour (warn + disable signing) or correct the comment.

### 3. `BLOXOS_UPDATE_PUBKEY` has incompatible meanings in the two binaries

- Hub `hub/update_signing.go:67` — the base64 **public key value**.
- Agent `agent/update_verify.go:66` — a filesystem **path** to the pinned key file.

Set once in a shared environment — precisely the single-box loopback deployment the transport gate preserves — the hub flips to offline mode while the agent attempts `os.ReadFile("MCowBQYDK2Vw…")`, gets ENOENT, and silently disables self-update. Suggest renaming the agent-side variable to `BLOXOS_UPDATE_PUBKEY_PATH` before this ships in installers.

### Smaller notes

- `hub/update_signing.go:221` — `if pub != nil && !ed25519.Verify(...)`. When `updateSigningPub` is nil (reachable via `initUpdateSigning:80-84`) the detached `.sig` is announced with no validation at all. Not exploitable today, but it is an unvalidated announce path.
- `agent/updater.go:58` — `SigAlg` is decoded and never read. Harmless while verification is unconditionally ed25519; an algorithm-confusion foothold the moment a second algorithm exists. Validate it or drop the field.
- `agent/update_verify.go:113` — `hex.DecodeString("")` succeeds, so `verifyUpdateSignature` accepts an empty SHA. Only the separate `version.SHA256 == ""` guard at `updater.go:69` prevents it. `authorizeUpdate` is documented as "the single gate every self-update must clear" but is not self-sufficient; add a `len == 64` check.
- `hub/agent_versions.go:207-232` and `hub/agentws.go:417-438` are copy-pasted binary-path candidate lists (one for hashing and locating `<binary>.sig`, one for serving). If they drift the hub signs one file and serves another — fail-closed at the agent, but a silent fleet-wide update stall. Nothing pins them together.

### Test gaps

- **No test exercises the gates-before-download ordering** — the PR's central guarantee. It holds only by reading `updater.go:95` against `:139`.
- `TestNoSignatureMeansNoAnnounce` (`hub/update_signing_test.go:336-343`) never calls `announceVersionToAgent`; it asserts `announcedSHAFor`/`announcedSignatureFor` only. The early return at `hub/agent_versions.go:278-284` that the test is named for is untested.
- `initUpdateSigning` and `loadOrGenerateUpdateSigningKey` are entirely untested — all env/path branches and every `log.Fatalf`.
- No test for the post-download SHA re-check in `performUpdate`.

### Verified as claimed

Transport gate and loopback exemption (literal only, no DNS resolution — correctly stricter); fail-closed on absent pinned key; OS bound into the signed message with byte-identical construction on both sides; both gates before download; detached-`.sig` precedence; and the `StartLimit*` change being a pure `[Service]`→`[Unit]` relocation with no behavioural difference ([systemd `6bf0f40`](https://github.com/systemd/systemd/commit/6bf0f408e4833152197fb38fb10a9989c89f3a59) keeps the `[Service]` spelling as a documented compatibility alias).

bokiko:
Reviewed at `54f75f7`. **The security property holds and I could not find a way around it.**

I went looking specifically for the failure mode where making `initUpdateSigning` non-fatal lets the hub boot with signing disabled and announce an *unsigned* update anyway — which would hand back the RCE this PR exists to close. It doesn't happen. Both gates fail closed:

- `hub/agent_versions.go:295-301` — no signature available, no announce.
- `hub/agent_versions.go:281-288` — agent hasn't reported `update_protocol >= 1` and `PUBLIC_URL` isn't TLS or loopback, no announce.

Also verified: `grep 'log.Fatal' hub/update_signing.go` returns nothing; `update_signing.go:143` is `case !os.IsNotExist(err)` returning `nil, ""`; the agent reads `BLOXOS_UPDATE_PUBKEY_PATH` (`agent/update_verify.go:84`) while the hub reads `BLOXOS_UPDATE_PUBKEY` (`hub/update_signing.go:76`), with no overlap left in non-test source.

## Non-blocking: every new fail-closed path is invisible to an operator

This PR adds three distinct ways for fleet updates to stop dead:

| Path | Site |
|---|---|
| signing key unreadable | `hub/update_signing.go:147` |
| explicit key missing | `hub/update_signing.go:153` |
| legacy agent on a plaintext hub | `hub/agent_versions.go:283` |

All three are a `log.Printf` and nothing else. `handleListVersions` returns `hub_sha`, `hub_short_sha`, `hub_mtime`, `hub_windows_sha`, `hub_windows_short_sha`, `hub_windows_mtime`, `agents`, `rollout_paused`, `pause_reason` — **none of which reports whether the hub can sign at all.**

In all three cases the dashboard shows `update_pending: true` on every agent indefinitely, which is indistinguishable from a rollout in progress. An operator whose fleet has silently stopped updating has no way to tell that from a fleet that is updating normally, short of reading hub stdout.

Suggested, since the state is already computed in `initUpdateSigning`: add `signing_enabled` and `signing_disabled_reason` to the `/api/versions` response, plus the per-machine legacy-withheld reason on each agent entry. Not a merge blocker — it breaks nothing and the class is already tracked — but this PR is where the three new paths are introduced, so it's the cheapest place to surface them.

## Merge ordering

`54f75f7` is 2 commits behind `main` and branch protection is `strict: true`, so this needs to be brought current before it can merge — and its green checks were run against `8a38f1b`. Worth waiting until the Dependabot queue stops moving rather than rebasing per merge.

Separately: `Agent Tests (windows)` is **not** in the required-checks list. This PR modifies `agent/updater_windows.go` and its Windows job is green, but that green is advisory — a Windows-only regression here would merge. That's a repo setting, not something this PR can fix.

bokiko:
Marking this PR as draft because signed agent updates remain blocked by the runtime shakedown tracked in #130.

Before this can return to review:

- complete the deployed-current-main shakedown;
- design and verify a migration that gives existing agents the pinned update key before they run a signature-enforcing binary;
- ensure refusing legacy agents cannot arm reconnect timers or trip the fleet-wide rollout circuit breaker;
- rebuild or rebase the work onto current `main`;
- rerun the complete agent, hub, Windows, race, and installer validation.

The current head must not be merged as-is. The branch is being preserved as reference work while those gates remain open.

bokiko:
Merged as f8f4714. Two known gaps disclosed in this PR's body are now tracked separately rather than left only in prose:

- #144 -- Windows applyPendingUpdate has no SHA re-check before swap (pre-existing, untouched by this PR, flagged by Codex during review)
- #145 -- signed announcements have no downgrade/monotonicity protection (deliberately deferred, per section 4 above)

Also closes the loop on #130's "do not start signed agent updates" line item -- see comment there.

PR #174, opened 2026-09-05
feat(hub+agent+ui): AI Sessions — read-only live visibility into AI coding tools

bokiko:
## Summary

Adds **AI Sessions**: read-only, live-only visibility into which supported AI coding tools (Claude Code, Codex, Kimi) are running on fleet machines. Two commits: the backend vertical slice (agent discovery, protocol, hub store, read API, admin switch) and the user-facing experience (SSE delivery and dashboard surfaces).

## User experience

- **Fleet pulse strip** gains an "AI Sessions" cell: running session count and how many machines they are on. Click-through to `/sessions`.
- **`/sessions`** lists live sessions grouped by machine. Each row shows the tool, the model with an **exact / inferred / unknown** mark, running duration, project folder name, and a coarse activity hint explicitly labelled *inferred*.
- **Machine page** gets an "AI Sessions" tab (with a count) on agent machines.
- **Settings** gets an admin-only "AI Sessions" tab with the switch, stating metadata-only, live-only, and the local opt-out.
- Header link and command-palette entry for navigation.
- **Running is the only definitive state**: a session is shown because its process exists. Idle is an inferred CPU reading and is never presented as "ended".
- States handled: loading, disabled by admin, empty, machine not reporting, error with retry, stale snapshot, and live-updates-paused. Sessions and machines disappear without a refresh when a process ends or an agent disconnects.
- Read-only throughout. No control acts on a session.

## Privacy contract

The wire contract lives in `proto/aisessions` and is shared by agent and hub so rules cannot drift. A session carries only: an opaque id (not the PID), tool, start time, and three `{value, source, confidence}` attributes:

- **project** — the working-directory **basename** only, withheld for the filesystem root, top-level directories, home directories, and any name equal to the OS username.
- **model** — only from an explicit `--model` (or Codex `-m`) flag among the tool's own arguments; a full id is `exact`, an alias like `opus` is `inferred`. Never read from the environment.
- **activity** — `active`/`idle` inferred from CPU time between scans, `unknown` on first observation. Never reported as exact.

**Never collected or transmitted:** prompts, responses, transcript contents, tool commands or output, environment variables, full argv, full paths, or OS usernames. **No history is retained**: the hub keeps one in-memory snapshot per connected machine and drops it when the socket unregisters, the machine is deleted, the switch is turned off, or 90s pass without a report. The hub re-sanitizes every inbound frame (whitelisted tools, capped counts and lengths, enumerated source/confidence, identifier-only model values, path-free project names); unknown fields are dropped. SSE payloads carry the same sanitized view and nothing else.

## Architecture

- **Agent** — `agent/aiscan` is a pure, cross-platform classifier and reducer: exact-token matching on the process name, `argv[0]`, or the script/module an interpreter runs (no substring matches, so `claude-desktop` or `codex-server` never match); wrapper scripts and self-spawned children collapse into one session; `python -m kimi_cli` is not misread as a model. `agent/ai_sessions.go` reads the process table via gopsutil (already a dependency; no new dependencies anywhere) and reports on the existing 30s cadence.
- **Hub** — snapshots are keyed by the socket's authenticated machine identity (the frame's own `machine_id` is ignored) and accepted only from the registered connection. `GET /api/ai-sessions` (fleet.read) and `PATCH /api/ai-sessions/settings` (fleet.admin). Switch persisted in a new `hub_settings` table; missing row = enabled, so no setup is required. SSE events on `/api/events`: `ai_sessions`, `ai_sessions_removed`, `ai_sessions_config`. The GET response includes the hub clock so clients age snapshots without clock agreement.
- **Dashboard** — the SSE provider gains a generic `subscribe(event, handler)` that re-attaches on reconnect; `AISessionsContext` is the single store, bootstrapping from GET on auth and every reconnect and applying SSE deltas. No second stream, no polling (the only timer is a 10s render clock for durations where sessions are displayed).
- **Backward compatibility** — older hubs log the unknown frame and continue; older agents never send it and treat the config frame as an id-less command they ignore. Windows builds vet and compile.

## Admin and local opt-outs

- **Admin switch** (default on): turning it off clears every snapshot and pushes an `ai_sessions_config` frame to all connected agents, which **stop scanning** immediately. The switch is cached with a monotonic revision captured atomically for every send; agents accept only strictly greater revisions on a connection, so a delayed registration "enabled" cannot undo a newer "disabled", and equal-revision duplicates or forgeries cannot flip the gate. Concurrent toggles are serialized under one lock so the persisted order matches revision order.
- **Agent gate**: scanning runs only after the hub has said "enabled" on the current connection, so an agent talking to an older hub never scans.
- **Local hard opt-out**: `BLOXOS_AI_SESSIONS=0` in the agent's environment; no hub setting can override it.

## Validation performed

- hub: `go vet ./...`, `go test -count=1 ./...`, `go test -race -count=1 -timeout=10m ./...` — pass, 0 data races.
- agent: `GOOS=linux` and `GOOS=windows` `go vet ./...`, `go vet -tags insecure`, test binaries compile for both targets; `agent/aiscan` tests pass with `-race`. Package-main agent tests compile for linux/windows and run in CI (they cannot execute on the macOS dev host).
- proto: `go test -race ./...` pass. `go mod tidy -diff` clean in all three modules.
- dashboard: `tsc --noEmit`, `eslint`, `next build` — pass.
- End-to-end on a scratch hub with a fake agent: enrollment, config frame with revisions, session reports, SSE `ai_sessions` and `ai_sessions_config` events observed, agent stopped and resumed on toggle.
- Test coverage: classifier positives/negatives; privacy regressions on agent, hub, and SSE with planted secrets, prompts, full paths, argv, env and unknown fields; stale expiry; spoofed and unregistered identity; displaced sockets; RBAC (viewer/operator/admin/unauthenticated); protocol compatibility (future schema, malformed frames, old-agent traffic); race safety; revision ordering and DB/cache consistency under concurrent toggles; removal and config broadcasts.
- Four existing hub tests were adjusted to tolerate the new post-registration config frame; nothing else in them changed.

## Residual risks

- The model field is shape-checked only (identifier charset, 64 chars); an identifier-shaped string from a compromised agent passes. Documented by a dedicated test.
- Activity is an uncalibrated CPU-rate heuristic (1%), always labelled inferred.
- A tool launched directly by a versioned binary path rather than its `claude` launcher is not classified.
- Hub-side stale eviction is lazy (on GET); the dashboard computes staleness itself, so a live-but-silent agent shows as stale rather than vanishing until the next bootstrap.
- Browser screenshots at desktop and phone widths were not taken during development; responsive and accessibility semantics (semantic lists, `time` elements, `role="switch"` with `aria-checked`, `role="status"` notices, text-plus-colour confidence marks, stacked phone layout) were verified by review and build only.
- The repo has no frontend test runner, so no component tests were added.

https://claude.ai/code/session_01DQHxs4yyRwNjaWmsEE76Tz

---

> [!NOTE]
> **Medium Risk**
> Adds process-table scanning and a new agent→hub metadata channel with strict sanitization, but misclassification or a compromised agent could still publish identifier-shaped model strings; admin toggles affect all connected agents.
> 
> **Overview**
> Introduces **AI Sessions**: read-only, live visibility into Claude Code, Codex, and Kimi processes on agent machines, bounded by a shared `proto/aisessions` contract and documented in **AGENTS.md**.
> 
> **Agent:** New `aiscan` classifier/reducer and `ai_sessions.go` scan the process table on the existing 30s metrics tick, gated by hub `ai_sessions_config` (revision-ordered) and local **`BLOXOS_AI_SESSIONS=0`**. Reports are sanitized before send; scan failures never tear down the connection.
> 
> **Hub:** In-memory snapshots per authenticated socket (no history), **`GET /api/ai-sessions`** / **`PATCH .../settings`** (persisted in new **`hub_settings`**, default on), re-sanitize on ingest, 90s stale expiry, SSE events **`ai_sessions`**, **`ai_sessions_removed`**, **`ai_sessions_config`**. Snapshots clear on disconnect, machine delete, or admin off.
> 
> **Dashboard:** **`AISessionsProvider`** bootstraps from GET on SSE reconnect; **`SSEContext.subscribe`** fans out named events without a second stream. Surfaces: fleet pulse cell, **`/sessions`**, machine **AI Sessions** tab, admin settings switch, nav/command palette.
> 
> Existing hub tests tolerate the post-registration config frame alongside **`agent_version`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86720d139481e5a56f26e5fc9db034389d8de847. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,6 +39,16 @@ Fleet management dashboard for AI machines. Go hub + agent, Next.js dashboard.
   trusted provisioning path.
 - The signature may come from a detached `<binary>.sig` produced offline, in
   which case the hub holds no private key — do not assume the hub signs.
+- **AI Sessions is metadata-only monitoring.** Agents report which supported
+  AI coding tools (Claude Code, Codex, Kimi) are running, reduced to the
+  contract in `proto/aisessions`: opaque id, tool, start time, project
+  **basename**, an explicitly-flagged model, and an inferred activity state,
+  each with its source and confidence. Never add prompts, responses,
+  transcripts, tool commands/output, environment variables, full argv, full
+  paths, or the OS username to that contract. The hub keeps live snapshots
+  only (no history) and re-sanitizes every frame. The admin switch defaults
+  to on; disabling it stops agent-side scanning via the `ai_sessions_config`
+  frame, and `BLOXOS_AI_SESSIONS=0` on an agent is a hard local opt-out.
 - Credentials NEVER committed — use env vars or local config
 - Dark mode is the default UI theme
 - Caddy TLS uses RSA-2048 keys, not ECDSA. Do not change without testing on stock Windows PowerShell 5.1.
--- a/agent/ai_sessions.go
+++ b/agent/ai_sessions.go
@@ -1,0 +1,137 @@
+package main
+
+// AI Sessions (checkpoint 1): read-only discovery of running AI coding
+// tools — Claude Code, Codex, Kimi — reported to the hub as metadata only.
+//
+// What leaves this machine is exactly aisessions.Session: an opaque id, the
+// tool name, start time, project basename, an explicitly-flagged model, and
+// a CPU-time activity inference. Classification and reduction live in the
+// platform-neutral aiscan package; this file only reads the live process
+// table (gopsutil, already a dependency) and sends the result. A monitored
+// process's environment is never read. Nothing is written or signalled.
+//
+// Scanning is gated (aiscan.Gate): it runs only after the hub has said the
+// feature is enabled on this connection, and never when the machine-local
+// BLOXOS_AI_SESSIONS opt-out is set.
+
+import (
+	"context"
+	"encoding/json"
+	"log"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/bokiko/bloxos/agent/aiscan"
+	"github.com/bokiko/bloxos/proto/aisessions"
+	"github.com/gorilla/websocket"
+	"github.com/shirou/gopsutil/v4/process"
+)
+
+// listAIProcesses reads the live process table via gopsutil (already a
+// dependency of the agent). It reads only the executable name for every
+// process and fetches the remaining metadata solely for candidates that
+// could be an AI tool, keeping the scan cheap on busy hosts.
+func listAIProcesses(ctx context.Context) ([]aiscan.Process, error) {
+	procs, err := process.ProcessesWithContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]aiscan.Process, 0, 8)
+	for _, p := range procs {
+		name, err := p.NameWithContext(ctx)
+		if err != nil {
+			continue
+		}
+		if !aiscan.Candidate(name) {
+			continue
+		}
+		argv, err := p.CmdlineSliceWithContext(ctx)
+		if err != nil {
+			argv = nil
+		}
+		if _, ok := aiscan.Classify(name, argv); !ok {
+			continue
+		}
+		ap := aiscan.Process{PID: p.Pid, Name: name, Argv: argv}
+		if ppid, err := p.PpidWithContext(ctx); err == nil {
+			ap.PPID = ppid
+		}
+		if cwd, err := p.CwdWithContext(ctx); err == nil {
+			ap.Cwd = cwd
+		}
+		if user, err := p.UsernameWithContext(ctx); err == nil {
+			ap.Username = user
+		}
+		if start, err := p.CreateTimeWithContext(ctx); err == nil {
+			ap.StartMS = start
+		}
+		if times, err := p.TimesWithContext(ctx); err == nil {
+			ap.CPUSeconds = times.User + times.System
+		} else {
+			ap.CPUSeconds = -1
+		}
+		out = append(out, ap)
+	}
+	return out, nil
+}
+
+var (
+	aiScanner = aiscan.NewScanner()
+	// aiGate holds the hub's runtime enable/disable decision for the
+	// current connection; runAgent resets it on every connect.
+	aiGate aiscan.Gate
+)
+
+// aiSessionsConfigType is the hub→agent frame carrying the admin switch.
+// Agents built before this feature route it through handleCommand, which
+// ignores any frame without a command id — so it is safe to send to every
+// agent.
+const aiSessionsConfigType = "ai_sessions_config"
+
+// handleAISessionsConfig applies an ai_sessions_config frame. When the hub
+// newly enables the feature, one report is sent right away rather than
+// waiting for the next 30s tick.
+func handleAISessionsConfig(conn *websocket.Conn, mu *sync.Mutex, machineID string, msg []byte) {
+	var cfg struct {
+		Enabled bool   `json:"enabled"`
+		Rev     uint64 `json:"rev"`
+	}
+	if err := json.Unmarshal(msg, &cfg); err != nil {
+		log.Printf("invalid %s frame: %v", aiSessionsConfigType, err)
+		return
+	}
+	if aiGate.Apply(cfg.Enabled, cfg.Rev) {
+		go sendAISessions(conn, mu, machineID)
+	}
+}
+
+// sendAISessions scans and reports. Failures are logged and never abort the
+// connection: this is auxiliary metadata, and a broken process table must
+// not take metrics or auto-update down with it.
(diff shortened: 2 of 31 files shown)


bokiko:
**Review: needs-work** (do not merge yet)

Feature design is strong: shared `proto/aisessions` contract, exact-token classifier, hub re-sanitize, socket-bound identity (frame `machine_id` ignored), revision-gated config, local `BLOXOS_AI_SESSIONS=0` hard opt-out, live-only in-memory store, read-only UI. CI is fully green (Hub, Hub race, Agent, Agent Windows, Dashboard).

Two P1s must land before ship. I independently verified both Codex findings.

### P1 — must fix

**1. Windows domain-qualified usernames leak as project names** (`proto/aisessions/aisessions.go` `ProjectFromDir`)

`Username` from gopsutil on Windows is often `DOMAIN\\alice` / `COMPUTER\\alice`. Comparison is `EqualFold(basename, username)` against the *full* string, so cwd `C:\\work\\alice` → project `alice` is not withheld. That ships an OS username as an “exact” project attribute — breaks the privacy contract on the happy path, not only under a compromised agent.

Fix: strip the domain/`\\` / `/` account component before comparing (and cover with a regression test).

**2. Disable race can re-publish snapshots and flip the dashboard “enabled”** (`hub/ai_sessions.go` `ingestAISessions` + `dashboard/.../AISessionsContext.tsx`)

`ingestAISessions` checks `aiSessionsEnabled()`, then later `putAt` + SSE `ai_sessions`. Concurrent `setAISessionsEnabled(false)` can clear + broadcast disabled, after which ingest still stores and broadcasts a fresh snapshot. Worse: the dashboard handler does `setEnabledState((e) => (e === false ? true : e ?? true))` on every `ai_sessions` event, so a late report can flip the UI back to enabled after an admin disable.

Fix: serialize ingest with the config lock (or re-check enabled+rev under the same lock before `putAt`/broadcast), and never infer `enabled=true` from a snapshot event — trust only `ai_sessions_config` / GET.

### P2 — should fix before or immediately after (freshness)

**3. In-flight GET can overwrite newer SSE state** (`AISessionsContext` `refreshSeqRef`)

`refreshSeq` only advances when a new GET starts. An `ai_sessions` / `_removed` that arrives after the response was generated but before `setMachines` still loses to the older snapshot. Sessions can blink out/in until the next report.

Invalidate outstanding bootstraps when SSE mutates state, or merge by hub `received_at`.

**4. Lazy stale eviction suppresses later removal SSE** (`aiSessionStore.live` + `removeAISessions`)

GET can silently delete a stale entry. A later disconnect/delete then finds nothing and emits no `ai_sessions_removed`, so other open dashboards keep the old sessions until reconnect. Broadcast removal on disconnect/delete regardless of store membership (or emit removal on lazy eviction).

### What’s already solid
- Identity binding + registered-connection check
- Sanitize whitelist/caps; exact-token matching (no `claude-desktop` false positives)
- Config revision gate on the agent (stale enabled cannot undo newer disabled)
- RBAC: GET `fleet.read`, PATCH `fleet.admin`
- Default-on with empty `hub_settings` row; disable clears store + pushes config
- Test surface looks serious (privacy planted secrets, RBAC, race, revision ordering)

### Bokiko call
- Default-on for process-table scanning fleet-wide: intentional product choice? (Local opt-out exists; still a privacy posture call.)
- Accept residual risk that a compromised agent can still send identifier-shaped model strings (documented) — fine if acknowledged.

### Verdict
**needs-work** — merge after P1 #1 and #2. P2 #3/#4 are strongly recommended in the same pass. Not a design block; do not merge as-is.

bokiko:
**P1/P2 follow-up (Engineering)** — pushed to `feature/ai-sessions`, head `86720d1`. **Not merging.**

| Commit | Fix |
|--------|-----|
| `fbc265c` | P1 username: `normalizeUsername()` for `DOMAIN\\user`, `COMPUTER\\user`, `user@DOMAIN` before basename compare |
| `7bc34ca` | P1 race: ingest holds config lock + re-checks enabled before put/SSE; dashboard no longer infers `enabled=true` from `ai_sessions` snapshots |
| `86720d1` | P2: bump `refreshSeqRef` on SSE (stale GET can't overwrite); `removeAISessions` always broadcasts removal |

**CI on `86720d1`:** Hub, Hub race, Agent, Agent (windows), Dashboard Lint+Build — all green. Bugbot neutral.

Still open product call: default-on process scanning. Awaiting bokiko/CoS merge call.

PR #195, opened 2026-09-07
fix: consistent backup and identity-preserving restore

bokiko:
## Summary

- Replace the unsafe live SQLite-only backup advice with a stopped-stack helper and clean, ownership-preserving restore procedure covering database/WAL, JWT/setup/signing identity and Caddy CA/config.
- Preserve original service running state on success/failure, refuse destination overwrite, keep private partial backups for diagnosis.
- Add five failure-path tests and a real network-isolated Compose restore smoke gate that checks every regular file's bytes/owner/mode and starts the restored initialized hub.
- Trigger Compose smoke on hub/agent/proto/dashboard changes. Require reusable full CI plus smoke on the exact release-tag commit before publishing images.
- Correct functional deployment/onboarding, 30-second telemetry, Windows terminal limitation and protocol-2 rollback-floor documentation. No visual redesign.

## Validation

- Five backup unit tests pass; Bash syntax and diff checks pass.
- actionlint v1.7.12 validates both changed workflows.
- Real disposable Linux Docker backup and fresh isolated restore passed: SQLite integrity; exact user, machine identity and durable credential rows; identical file-backed identity and CA; restored files UID/GID65532 mode600; healthy hub with needs_setup=false.
- New compose-backup.py smoke executed successfully against that disposable initialized stack; exact archive contents/ownership preserved, no network access for restored hub, only its uniquely created temporary volumes cleaned up. Backup retained.
- No production fleet deployment, agent update, live volume deletion or published release.

## Review / scope

Unit F of the approved reliability plan (finding3 and functional release/integration documentation). Independent engineer review and exact-head CI must complete before merge. Local Qwen/DeepSeek were consulted; their archive-path/ownership claims were contradicted by the real restore, so no unsafe root extraction or ownership weakening was adopted.

---

> [!NOTE]
> **Medium Risk**
> Backup stops hub/Caddy briefly and touches identity-critical state; release publishing now depends on full CI at the tag, which changes the release path but reduces shipping untested tags.
> 
> **Overview**
> Replaces “copy `bloxos.db`” guidance with a **stopped-stack Compose backup** (`scripts/backup-compose.sh`) that archives hub/Caddy data and config, uses a Docker lock to block overlapping runs, atomically publishes `SHA256SUMS`, and restores prior container running state on success or failure. **`docs/backup-restore.md`** documents backup, checksum verification, and clean isolated restore.
> 
> **CI/release:** `ci.yml` is **`workflow_call`**-able and the hub job runs **`scripts/test_backup_compose.py`** (fake Docker failure-path tests). Compose smoke (`compose-enroll.sh`) now runs **`compose-backup.py`** for a network-isolated restore check. **`docker.yml`** widens PR path filters, adds **`verify-release`** (full CI on `v*` tags), and blocks image **publish** until smoke and that suite pass on the tag commit.
> 
> **Docs:** README and `docker/README.md` promote Compose quick start, WAL-safe backup links, telemetry cadence, Windows terminal limits, and protocol-2 update-floor behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54af271f056fde3b3f8ad55a39842deb02bfa440. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
       - main
   pull_request:
   workflow_dispatch:
+  workflow_call:
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.ref }}
@@ -56,6 +57,9 @@ jobs:
 
       - name: Run hub tests
         run: go test -count=1 ./...
+
+      - name: Test backup failure and restart handling
+        run: python3 ../scripts/test_backup_compose.py
 
   # A real gate as of issue #60: the hub's DB handle and agent registry now live
   # on a Server instance, so tests build isolated servers instead of reassigning
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,7 +8,13 @@ on:
       - ".dockerignore"
       - "docker/**"
       - "scripts/smoke/**"
+      - "scripts/backup-compose.sh"
+      - "hub/**"
+      - "agent/**"
+      - "proto/**"
+      - "dashboard/**"
       - ".github/workflows/docker.yml"
+      - ".github/workflows/ci.yml"
   push:
     tags: ["v*"]
   workflow_dispatch:
@@ -25,6 +31,13 @@ env:
   DASHBOARD_IMAGE: ghcr.io/${{ github.repository_owner }}/bloxos-dashboard
 
 jobs:
+  # A release tag must pass the same full suite as a PR, at the tag's exact
+  # commit. A green older main/PR run does not validate a new tagged tree.
+  verify-release:
+    name: Release verification
+    if: startsWith(github.ref, 'refs/tags/v')
+    uses: ./.github/workflows/ci.yml
+
   # Builds both images for the runner's platform and enrolls the runner as an
   # agent through the Compose stack (scripts/smoke/compose-enroll.sh). Gates
   # publishing and checks container-related pull requests.
@@ -65,7 +78,7 @@ jobs:
   # behind the default tags.
   publish:
     name: Publish images
-    needs: smoke
+    needs: [smoke, verify-release]
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     permissions:
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ If you've ever opened five browser tabs to check on five machines, this is for y
 ## What you actually get
 
 ### Live, not polled
-Metrics stream over WebSocket from agent to hub, then push to your browser via SSE. CPU, RAM, GPU usage, GPU power, thermals, network throughput, disk I/O — all updating multiple times per second. No 15-second scrape intervals. No "did it just freeze or is it just slow." When a value changes, you see it change.
+Metrics stream over WebSocket from agent to hub, then push to your browser via SSE. The normal agent snapshot cadence is 30 seconds, with on-demand refresh. Freshness indicators distinguish current readings from stale data. Power history samples component sensors locally every second and uploads 30-second averages and sampled peaks; this is not a wall-power meter.
 
 ### Real hardware inventory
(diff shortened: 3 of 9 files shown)


bokiko (commented):
## Request changes

**Head:** `fe7bb38`

(GitHub blocks formal REQUEST_CHANGES on this account for this PR — **verdict is still request-changes**.)

No P1. Overall design is right: stop hub+caddy, copy full `/data`+Caddy state (not live `bloxos.db` alone), restart only previously running services, refuse destination overwrite, keep private incomplete dirs, restore into a fresh isolated project with ownership preserved. Docs correctly warn against live extract, volume delete, and minting a new CA. Tag-only `verify-release` + `publish` gating looks correct; Compose smoke on this PR already green (includes the new backup restore gate).

### P2 — `SHA256SUMS` can look valid while incomplete (`scripts/backup-compose.sh`)
Docs treat presence of `SHA256SUMS` as “complete backup,” but the script redirects `shasum` straight into that path. An interrupt, tool failure, or truncated write can leave an empty/partial manifest that `shasum -c` still accepts for whatever lines exist. **Write to a temp file, verify the expected five entries, then atomically rename to `SHA256SUMS`.** Add a failure-path test.

### P2 — destination path is not race-safe after `mkdir`
`mkdir` rejects an existing final path, then later writes go through that pathname. Between create and the `compose config` / `docker cp` redirections, another process can replace the directory with a symlink and divert private key material. Prefer an exclusive temp dir + atomic finalize (or document/enforce exclusive parent ownership).

### P2 — concurrent backups can succeed with inconsistent data
No per-Compose-project lock. Backup A stops services; B observes them stopped and skips restart tracking; A’s EXIT trap restarts while B is still copying → B can capture live-changing SQLite/WAL/Caddy bytes and still emit success. Add a project-scoped lock and re-verify stopped state immediately before copy.

### P3 (non-blocking)
Unit fake Docker doesn’t cover stop/restart/`shasum`/multi-ID/symlink/concurrent failures; smoke doesn’t run `shasum -c`, assert full manifest, or start restored Caddy. Tighten when convenient.

### CI
Hub / Agent / Shared Protocol / Dashboard / **Compose smoke** green. Hub Tests (race) still running at review time.

Not merging.

bokiko:
Addressed the three review blockers in the new head:

- Checksum generation writes `.pending`, validates exactly the five expected filenames and 64-digit SHA values, verifies checksums, then atomically renames to `SHA256SUMS`. A fake successful-but-partial checksum writer now fails without publishing a completion manifest.
- Backup parent must be owned by the invoking OS user and not group/world writable, before creating the private destination. Existing files/symlinks remain refused. Same-user hostile processes are explicitly outside this local file boundary.
- A uniquely named, stopped, network-less Docker lock container excludes concurrent clients on the same daemon before recording service state. It never executes the hub; normal cleanup removes only the owned lock and its anonymous volume. Stale-lock recovery is documented. This also works with remote Docker clients, unlike a host-local lock.

Eight unit tests pass. Re-ran real isolated Docker smoke: complete manifest checks pass, a competing backup is refused without stopping the source, all archive bytes/UID/GID/modes survive restore and the initialized hub is healthy. Its uniquely created test containers/volumes were cleaned; backup retained. Exact-head CI is running again; not merging on the earlier green run.

bokiko (commented):
## Re-review: approve

**Head:** `54af271`

(GitHub blocks formal APPROVE on this account for this PR — **verdict is still approve**.)

### Prior P2s closed
1. **Atomic `SHA256SUMS`:** writes `SHA256SUMS.pending`, validates exactly five expected names + 64-hex digests, runs `shasum -c`, then `mv` to `SHA256SUMS`. Incomplete never published as the final manifest. Unit test covers a lying checksum tool.
2. **Destination TOCTOU:** parent must be owned by the invoking user and not group/world-writable before `mkdir`; docs steer off `/tmp`. Same-user hostile processes correctly called out of scope.
3. **Concurrent backups:** exclusive stopped Docker lock container `bloxos-backup-lock-<hub-id>` acquired **before** observing running state / stop / copy; overlapping create fails; EXIT always tries `docker rm -v`. Unit + smoke cover refusal and source staying up.

### Residual (non-blocking)
- Stale lock after hard kill needs manual `docker rm -v` of that exact lock container (documented).
- Manual restore still depends on operators verifying project/volume identity.

### CI
Agent Tests green on this head when reviewed; Hub / race / Dashboard / Compose smoke still settling — prior-head smoke was green and the new smoke asserts include checksum + concurrent lock.

Not merging.

PR #196, opened 2026-09-07
hub: lifecycle safety — terminal single-close, committed-identity ingestion, atomic delete, stream revocation

bokiko:
Unit A of the reliability plan (REPORT.md findings #11, #12, #1, #2). Hub only; the dashboard consumer of `machine_removed` is Unit D.

## What changed

- **Terminal relay single-close** (`hub/terminal.go`): the inactivity timer and both reader goroutines ended the relay through a check-then-close select on one channel; two could both see it open and the second close panicked the process. One shared `sync.OnceFunc` now ends the relay.
- **Committed-identity ingestion** (`hub/agentws.go`): services, containers and hardware_info fell back to the payload's `machine_id` before the first metrics frame established identity, so a valid install-token socket could rewrite another machine's inventory without enrolling. Inventory writes now require a committed, registered identity. A fresh enrollment's hardware snapshot (sent right after its first metrics, before `enrollment_committed`) is held and applied at the commit point, so first enrollment keeps its inventory.
- **Atomic machine deletion** (`hub/main.go`, `hub/agentws.go`, `hub/server.go`, `hub/agent_versions.go`): delete removed rows and the registry entry but never closed the socket; the next metrics frame recreated the row as online with no credential and no command routing. Delete now takes the same auth writer lock as credential revocation (no handshake can validate or register meanwhile) plus an ingestion barrier: every persisting frame handler runs under its read side with the registration check inside, and delete holds the write side across registry removal, socket close and row deletion. A frame is either fully persisted before the delete or refused after it, for the registered socket, a displaced older socket still mid-frame, and any retry. Broadcasts sit inside the same barrier section so the additive `machine_removed` `{machine_id}` SSE event is always the last event for that machine, and the update announce arms its reconnect expectation only through the barrier. Running-version and reconnect-expectation bookkeeping is cleared on delete. Lock order everywhere: `agentAuthMu`, then `ingestMu`, then `agentsMu`.
- **Event stream revocation** (`hub/main.go`, `hub/auth.go`, `hub/rbac.go`): `/api/events` checked authorization once at the handshake, so an expired token or a deleted user kept receiving fleet telemetry for as long as the connection was held. The stream now enforces the token's `exp` with its own timer and re-checks user existence on a 60 s ticker, failing closed on lookup errors.

## Regression tests (`hub/lifecycle_safety_test.go`, 17 tests)

Concurrent terminal teardown under repetition; pre-identity foreign frames rejected while first enrollment still stores hardware at commit; delete while connected (socket closed, no resurrection from a late frame, rollout state cleared, `machine_removed` emitted); delete blocked behind a credential-validated handshake paused before registration; delete blocked behind a frame parked inside the barrier; a displaced socket mid-frame refused and the delete still wins; removal event strictly last; announce cannot re-arm a reconnect expectation after delete; stream ends on user deletion, on the periodic check, and at the exact token expiry.

## Verification

- `go vet` clean; targeted lifecycle plus enrollment, identity, rollout, and power-history tests pass.
- Full hub suite with `-race` passed locally on the pre-barrier tree (468 s); the run on this exact tree is in progress locally, and the required CI race job is the gate.
- No agent, proto, dashboard, or script changes. No deployment.

---

> [!NOTE]
> **High Risk**
> Changes core agent ingestion, machine deletion locking, and long-lived SSE auth in concurrency-sensitive paths; mistakes could drop live agents, stall deletes, or leak telemetry after user removal.
> 
> **Overview**
> This PR hardens hub lifecycle and realtime paths so deleted or displaced agents cannot resurrect fleet state, and long-lived dashboard streams fail closed when auth is no longer valid.
> 
> **Machine deletion** is now coordinated with agent auth and ingestion: delete holds `agentAuthMu` (like credential revoke) so no new handshake can register mid-removal, and `ingestMu` write lock so in-flight frames finish or are refused before rows are removed. The live socket is closed, rollout reconnect expectations and `/api/versions` entries are cleared, and dashboards get a **`machine_removed`** control SSE event that is ordered after any final in-barrier metrics broadcast.
> 
> **Agent WebSocket ingestion** routes metrics, services, containers, hardware, version reports, AI sessions, and power commits through **`ingestFrame`**, which re-checks registry ownership under the barrier. Uncommitted enrollments no longer persist inventory or version data (held until **`enrollment_committed`**); pre-identity frames cannot target another machine’s `machine_id`. **Power history** commits under the barrier but sends ACKs outside it so a blocked peer cannot stall delete.
> 
> **Rollout**: reconnect expectations after version announce arm only via **`armReconnectIfRegistered`** inside the same barrier, avoiding phantom failures after delete.
> 
> **SSE (`/api/events`)** now carries JWT **`exp`** on context, closes streams on exact expiry and periodic user revalidation, and **`broadcastSSEControl`** kicks slow clients that would drop **`machine_removed`** instead of delivering it.
> 
> **Terminal relay** uses a single **`sync.OnceFunc`** to close the shared done channel, fixing concurrent teardown panics.
> 
> Extensive regression tests cover delete/resurrection, handshake/barrier interleaving, SSE revocation, enrollment holding, and power ACK not blocking delete.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf356453cdc7dfd4e0244d9b12805153e3423073. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/hub/agent_versions.go
+++ b/hub/agent_versions.go
@@ -359,7 +359,16 @@ func (s *Server) announceVersionToAgent(machineID string, agent *ConnectedAgent)
 		return
 	}
 
-	expectReconnect(machineID)
+	s.armReconnectIfRegistered(machineID, agent)
+}
+
+// armReconnectIfRegistered arms the reconnect expectation for an announced
+// update only while agent still owns the registry entry, under the ingestion
+// barrier. The announce runs on its own goroutine, so without this a delete
+// that already cleared the expectation could be followed by a late arm and a
+// phantom rollout failure for a machine that no longer exists.
+func (s *Server) armReconnectIfRegistered(machineID string, agent *ConnectedAgent) bool {
+	return s.ingestFrame(machineID, agent, true, func() { expectReconnect(machineID) })
 }
 
 // announceDecision is the single place that decides whether an update may be
@@ -610,6 +619,15 @@ func expectReconnect(machineID string) {
 	pendingReconnectsMu.Lock()
 	defer pendingReconnectsMu.Unlock()
 	pendingReconnects[machineID] = time.Now().Add(reconnectExpectation)
+}
+
+// forgetAgentVersion drops the last-reported version for a machine that no
+// longer exists, so /api/versions stops listing it and no later reconnect
+// expectation can record a phantom rollout failure for it.
+func forgetAgentVersion(machineID string) {
+	agentRunningVersionsMu.Lock()
+	delete(agentRunningVersions, machineID)
+	agentRunningVersionsMu.Unlock()
 }
 
 func clearReconnectExpectation(machineID string) {
--- a/hub/agentws.go
+++ b/hub/agentws.go
@@ -870,6 +870,10 @@ func (s *Server) handleAgentWS(c echo.Context) error {
 	// Hold it and apply it at the commit point so first enrollment keeps
 	// its inventory without ever writing under an uncommitted identity.
 	var pendingHardwareInfo []byte
+	// Same for the agent's version report: it is sent once per connect and
+	// would otherwise create a /api/versions entry for an enrollment that may
+	// never commit.
+	var pendingVersionReport *agentVersionReport
 	defer func() {
 		s.unregisterAgentConnection(machineID, agent)
 		// A fresh enrollment that never committed was never registered, so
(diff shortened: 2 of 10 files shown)


bokiko (commented):
## Request changes

**Head:** `9c14b9f`

(GitHub blocks formal REQUEST_CHANGES on this account for this PR — **verdict is still request-changes**.)

No P1. The barrier design (agentAuthMu → ingestMu → agentsMu), inventory identity binding, hardware hold until commit, terminal `OnceFunc`, SSE exp/revalidate fail-closed, and `armReconnectIfRegistered` look sound. Lifecycle tests cover the advertised handshake/ingest/SSE/terminal paths.

### P2 — late metrics can repopulate `machineLatency` after delete
Latency is written **before** `ingestFrame` (`hub/agentws.go` ~1135–1137); delete clears it (`hub/main.go` ~848–850). A late refused frame can still leave latency state. Move the mutation inside the barrier callback after the registration check; add a regression.

### P2 — uncommitted enrollment can poison `agentRunningVersions`
`agent_running_version` uses `ingestFrame(..., registered, ...)` and the barrier skips the registration check when `registered == false`. A fresh enrollment can record version inventory before `enrollment_committed`; abort/disconnect may leave a phantom entry for `/api/versions`. Gate on committed registration or clean up on aborted enrollment; add a test.

### P2 — `machine_removed` can be silently dropped
Same lossy 64-entry SSE queue as telemetry (`broadcastSSE` default-drop). Under a full client buffer the deletion event can vanish, leaving the dashboard card until snapshot. Prioritize/reliably deliver control events (or force snapshot invalidation) and test a full queue.

### P3
Targeted re-enrollment vs delete not covered; dashboard consumer of `machine_removed` deferred to Unit D (noted).

Not merging.

bokiko:
Root re-review: the three initial P2s are fixed by committed-ingestion, latency inside the barrier, and full-queue control invalidation. The additional blocked-ACK defect is fixed in 670a04f: commit returns ACK bytes without network I/O, then ACK is written outside the barrier with a write deadline. Added cf35645 regression using the real WebSocket read loop: hold the agent write mutex, send a power batch, observe its durable commit, and require DELETE to return before releasing the writer. Targeted race tests pass. Waiting for fresh full exact-head CI before merge.

bokiko (commented):
Approve at `cf35645` (COMMENT — leave merge to bokiko).

Re-review of prior request-changes at `9c14b9f` plus the follow-up commits:

- **P2 latency-before-barrier — fixed.** Metrics path writes `machineLatency` inside the same `ingestFrame` section as upsert/store/broadcast. `TestDisplacedSocketMidWriteCannotResurrect` asserts the cache is not repopulated after delete.
- **P2 pre-commit `agent_running_version` phantom — fixed.** Unregistered fresh enrollments hold `pendingVersionReport` until `enrollment_committed`; aborted enrollments leave no `/api/versions` entry. Covered by `TestFreshEnrollmentVersionReportHeldUntilCommit`.
- **P2 lossy `machine_removed` — fixed.** Delete uses `broadcastSSEControl` (kick full-queue subscribers instead of dropping). Covered by `TestMachineRemovedKicksOverflowedStream` and `TestSSEStreamEndsOnControlOverflow` (real unread stream).
- **Blocked power ACK — fixed (`670a04f` / `cf35645`).** `commitPowerHistoryFrame` does DB only under the barrier; the read loop sends ACK afterward under `WriteMu` + 10s deadline. `TestPowerCommitUnderBarrierPerformsNoSocketWrite` and `TestDeleteDoesNotWaitForBlockedPowerACK` (holds `WriteMu`, delete still completes) lock it in.

Original Unit A behavior (terminal `OnceFunc`, committed-identity ingestion, delete barrier + socket close, SSE exp/user recheck) still looks solid.

**Nit (non-blocking):** `ingestPowerHistory` still exists as a test helper that writes ACK itself; production WS path correctly uses commit+external ACK — keep it that way.

CI: Agent/Windows/Dashboard/Shared Protocol green at review time; Hub / Hub race / Compose still running — confirm green before merge. After this lands, `#199` / `#201` can retarget/rebase.

PR #197, opened 2026-09-07
fix: reliable alert incidents, duration and bounded delivery

bokiko:
## Summary

Unit C of the approved reliability review (finding16):
- Treat active + acknowledged rows as one open incident, so acknowledgement does not re-notify the same breach. Resolve legacy duplicate rows together on recovery; retain ongoing acknowledged incidents during cleanup.
- Honor configured duration using small per-server observed-continuity state, reset on healthy/missing/stale readings, rule changes, gaps or restart. No new persistence subsystem.
- Preserve NULL-versus-zero for CPU/RAM/disk; do not use offline/stale metrics as recovery. Avoid inserting alerts for machines deleted during evaluation.
- Apply DB/SSE transitions before bounded Telegram delivery: 5s/request, 10s/batch, bounded body reads, no token-bearing URL error logging, no unbounded goroutines or retry queue.

## Validation

- Existing alert regressions plus seven new tests pass with race detector (ack/recovery/new incident; duration reset; missing/offline/stale readings; legacy duplicates+cleanup; partial RAM/disk; fake HTTP header/body stalls).
- `go vet ./...` and diff checks pass.
- Claude independently reviewed source/tests/docs and found no blockers. Qwen local also consulted; its pending-map claim is disproven by fresh-map construction each evaluation and the passing interruption regression.
- No real Telegram messages, live fleet changes or database migration.
- Full CI and exact-head review remain required before merge. Separate PR, not bundled with hub lifecycle/onboarding/dashboard work.

---

> [!NOTE]
> **Medium Risk**
> Changes core alert firing, resolution, and notification timing across all fleets; behavior is well-tested but mis-tuned freshness/duration rules could still miss or delay real incidents.
> 
> **Overview**
> Hardens **alert evaluation** so operators get consistent incidents instead of false fires, duplicate notifications, or hung eval loops.
> 
> **Active and acknowledged** rows are treated as one open incident: ack suppresses re-notification for the same breach, recovery resolves both, and legacy duplicate rows are updated/resolved together. **Cleanup** now deletes only **resolved** history older than 30 days, so ongoing acknowledged incidents are kept.
> 
> **`duration_secs`** uses in-memory continuity tracking (serialized via `alertEvalMu`): the timer resets on healthy readings, missing/stale sensors, rule changes, long evaluation gaps, or hub restart; API-polled machines get freshness windows based on poll interval so normal poll gaps do not break duration or falsely trigger offline rules.
> 
> Metric evaluation stops **COALESCE-to-zero** for CPU/RAM/disk: NULL/invalid readings are skipped (not recovery, not low-threshold triggers), threshold rules require **online** machines with **fresh** samples, and alert inserts are guarded when a machine is deleted mid-pass. Failed prefetch of open alerts **aborts** the pass instead of assuming no incident.
> 
> **Telegram** is deferred until after DB/SSE work: messages are sent in a **batched** pass with a **5s per-request** and **10s total** budget, bounded body reads, and safer logging (no token-bearing URL errors). **`docs/alerts.md`** documents the lifecycle; **`alert_lifecycle_test.go`** adds regression coverage for ack, duration, missing data, API cadence, legacy duplicates, and delivery timeouts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5ceb05f9e138e473abef344c79017855af46c35. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko:
Fixed the API polling-cadence regression in a508887 (current head 8ebb612 also brings in main). Native freshness remains 120 seconds; enabled API machines allow their configured interval plus one 30-second evaluation tick. Regression test uses a 300-second poll and 180-second duration: expected gaps do not reset the condition, while an overdue healthy-looking sample cannot resolve an incident. Focused race tests passed. Fresh full CI and Compose are running on this exact head.

bokiko (commented):
Request changes at `8ebb612`.

**P1 — early returns wipe all duration pending state.** `evaluateAlertsAt` always runs `defer func() { s.alertPending = nextPending }()` with `nextPending` starting empty. Any failure before the rule loop finishes (`getAlertRules`, machines query, `rows.Err`, unresolved-alerts query/`Err`) returns and **replaces** `s.alertPending` with `{}`. A transient DB blip silently resets every in-flight `duration_secs` timer, delaying legitimate alerts. Preserve the previous map on abort (e.g. only assign `nextPending` on success, or `committed` flag).

**P2 — Telegram batch runs while holding `alertEvalMu`.** Defers are LIFO: `sendTelegramBatch` runs **before** `Unlock`, so a configured bot can hold the eval mutex for up to the 10s batch budget. That stalls the 30s eval loop (and any other `alertEvalMu` waiter) on notification delivery. Collect messages under the lock, unlock, then deliver — matching the PR’s “DB/SSE before Telegram” intent without serializing HTTP on the mutex.

**P3 — `pending.rule != rule` compares the full `AlertRule` including `CreatedAt`/`Name`.** Rule edits correctly reset duration; ensure `created_at` string stability from SQLite cannot flap the comparison. Prefer comparing the fields that define the condition (`metric`/`operator`/`threshold`/`duration_secs`) or a content hash.

**Looks correct:** active+acknowledged as one incident; recovery resolves duplicates; cleanup only deletes `resolved`; NULL CPU/RAM/disk skipped (not zero/recovery); offline/stale metric paths do not resolve; GPU 0 skip retained; deleted-machine `INSERT … WHERE EXISTS`; API poll freshness `poll_interval+30` when `>90` (native 120s); `TestAlertDurationHonorsSlowAPIPollCadence` (300s/180s); Telegram timeouts, bounded body read, no token URL in logs; docs/alerts.md matches the intent.

**CI:** Dashboard / Agent / Windows / Shared Protocol green at review time; Hub / Hub race / Compose still running. Bugbot neutral.

Do not merge until the P1 pending-wipe on error paths is fixed.

bokiko:
Addressed the latest review in 3599b52. Pending-map replacement is now registered only after successful database prefetch, so aborted reads preserve the last observation; the existing 90-second gap bound and missing-sensor reset still apply. Added real SQLite table-unavailability regression proving preserved timing and subsequent firing. getAlertRules now returns rows.Err. Defer ordering now releases alertEvalMu before bounded Telegram delivery. Focused alert/Telegram race tests passed in 3.3s; exact-head CI is rerunning.

bokiko (commented):
Approve at `3599b52` (COMMENT — leave merge to bokiko).

Re-review of the prior request-changes:

- **P1 fixed:** `nextPending` + its assign-defer are registered only after rules/machines/unresolved-alerts prefetch succeeds. Aborted DB reads return without replacing `alertPending`. Covered by `TestAlertAbortedReadPreservesPendingDuration` (rename `alerts` mid-flight).
- **P2 fixed:** `sendTelegramBatch` is deferred *before* `Lock`, so LIFO runs Telegram **after** `Unlock`. Docs updated accordingly.
- **`getAlertRules`:** now returns `rows.Err()` — good.

Residual nits only (non-blocking): full `AlertRule` struct equality for pending reset (including `CreatedAt`/`Name`); same as before.

Prior lifecycle behavior (ack=open incident, NULL≠zero, API poll freshness, bounded Telegram, cleanup only `resolved`) still looks solid.

CI: Agent/Dashboard/Shared Protocol green at review time; Hub / Hub race / Compose still running — confirm green before merge.

bokiko:
One additional regression found during integrated cadence validation, fixing before merge: the pre-existing offline rule counts a normal 300-second API poll gap as downtime and fires at 180 seconds. New regression fails on current head (normal scheduled poll gap generated an offline incident). Applying API-only overdue-time accounting: elapsed since last success minus configured interval, bounded at zero; native-agent semantics unchanged. Test also checks overdue firing and recovery. This is still Unit C, not a new architecture or user setup requirement.

bokiko:
API offline-cadence fix is now f03c20e. Regression proved the previous behavior (offline alert at 180s during a configured 300s gap); it now passes, including alert after 421s and resolution on a successful poll. Offline time for API machines is max(0, elapsed - configured interval); native behavior unchanged. All focused alert/Telegram race tests pass (3.6s). Fresh exact-head CI running.

bokiko (commented):
Approve at `f03c20e` (COMMENT — leave merge to bokiko).

Re-review of the API offline-cadence fix on top of the already-approved `3599b52` base:

- **`machine_offline` for API pollers:** `offlineSecs = max(0, age - poll_interval)` when `pollInterval` is in `[30, 3600]`. A normal 300s poll gap no longer trips the default 120s offline rule; fire only once the next poll is overdue (`421s` → 121s overdue). Native agents (`pollInterval` 0) unchanged.
- **Regression:** `TestAlertOfflineDoesNotCountScheduledAPIPollGap` covers no-alert at 180s, alert at 421s, resolve after successful poll at 450s.
- **Docs** updated to match.
- Prior fixes still present: pending map only after prefetch; Telegram after `Unlock`; `getAlertRules` returns `rows.Err()`.

**Nit (non-blocking):** the alert message reports the *adjusted* overdue seconds (not wall-clock since last_seen), which matches the threshold math but can look shorter than wall age.

CI: Agent/Dashboard/Shared Protocol green at review time; Hub / Hub race / Compose / Windows still running — confirm green before merge.

PR #15, opened 2026-04-05
KyZN: fix LOW findings (20260405-112119-bf349ae6)

bokiko:
## Analysis Fixes

Applied fixes for findings at severity **LOW** and above.

**Run ID:** `20260405-112119-bf349ae6`
**Cost:** $0.68
**Batches:** 1 applied (LOW), 3 failed, 0 skipped
**Diff:** 435 lines

### Changes
```
 SECURITY-AUDIT.md                        | 327 +++++++++++++++++++++++++++++++
 apps/api/src/index.ts                    |   5 +-
 apps/api/src/middleware/authorization.ts |   6 +-
 apps/api/src/middleware/csrf.ts          |  23 ++-
 apps/api/src/routes/auth.ts              |   4 +-
 apps/api/src/routes/rigs.ts              |  19 +-
 apps/api/src/routes/users.ts             |   9 +-
 apps/api/src/services/alert-service.ts   |  19 ++
 apps/api/src/services/auth-service.ts    |   2 +-
 apps/api/src/services/gpu-poller.ts      |   8 +-
 apps/api/src/services/ssh-manager.ts     |  13 +-
 11 files changed, 406 insertions(+), 29 deletions(-)
```

### What Was Fixed

**LOW:** The 1 remaining error is `routes/alerts.ts(171,42)` — a pre-existing error in a file I never touched. All 4 fixes are complete.

---

## Summary

| Finding | Change | File |
|---------|--------|------|
| **SEC-007** | CSRF now enabled by default; disabled only when `DISABLE_CSRF=true` or `NODE_ENV=test` | `apps/api/src/index.ts:231` |
| **PERF-006** | Merged `prisma.rig.update` + `prisma.rig.findUnique` into single `update(..., include: { groups: true })` | `apps/api/src/routes/rigs.ts:402` |
| **PERF-007** | Added `limit`/`offset` query params (default 50, max 200) to `GET /api/rigs` via `take`/`skip` | `apps/api/src/routes/rigs.ts:38` |
| **PERF-008** | Added `clearStaleEntries()` to `AlertService` — removes deleted rig IDs from `previousHashrates` and `lastAlertTimes` maps; called at the start of `checkOfflineRigs()` each cycle | `apps/api/src/services/alert-service.ts:229` |

The 1 remaining hook warning (`alerts.ts:171`) is a pre-existing error that predates this branch.

### Approach
Findings were batched by severity tier (CRITICAL → HIGH → MEDIUM → LOW).
[...]

PR #199, opened 2026-09-07
hub: onboarding repair — offset-safe token cleanup, retryable setup, normalized PUBLIC_URL, root without sudo

bokiko:
Unit B of the reliability plan (REPORT.md findings #14, #15, #17). **Stacked on #196**: base is `fix/hub-lifecycle-safety`; retarget to `main` once #196 merges. Hub only.

## What changed

- **Offset-safe token cleanup** (`hub/agentws.go`, `hub/main.go`): rows written by older hubs hold the driver's rendering of a `time.Time` (`2026-09-07 17:45:51 -0400 EDT`), which SQLite's date functions cannot parse, and the old `expires_at < datetime('now')` text comparison mis-ordered by offset (fresh tokens deleted west of UTC, stale ones kept east of it). New rows store UTC RFC3339 text, readable by Go and SQLite. Cleanup now decides in Go with `parseTokenExpiry`, which accepts every form the column has held, and judges legacy rows by their actual instant. Unparseable rows are left alone and logged.
- **Retryable setup with the single-admin guarantee** (`hub/auth.go`): the token is checked first, consumed atomically only after validation and hashing succeed, and restored if the insert fails. Concurrent valid requests still produce exactly one admin.
- **Normalized PUBLIC_URL** (`hub/agentws.go`, `hub/join.go`): trailing slashes are stripped once in `publicAndWebsocketBase`, so no `//install.sh` paths and a clean mint-time binding; the join drift check compares origins so tokens minted by older hubs with the slash stored still serve. Mint-time byte binding and pinned-CA trust are unchanged.
- **Root without sudo** (`hub/install_commands.go`, embedded `install.sh`): both scripts define `SUDO` from `id -u` and use it for every privileged step. Existing tests that asserted the literal `sudo` now assert `$SUDO`.

## Tests (`hub/onboarding_repair_test.go`, 7 tests)

Legacy offset rows inserted the way older hubs wrote them (fresh west, expired east and west, unnamed zone) plus UTC text, judged by instant and still validating; minted rows stored as UTC text SQLite can read; setup token survives a validation failure and an injected storage failure then succeeds once and is not reusable; eight concurrent setups create one admin; trailing-slash `PUBLIC_URL` yields clean commands and binding; legacy slash binding still joins; both scripts define the shim before any privileged step and never call `sudo` directly.

## Verification

`go vet` clean; new tests plus all join, installer, enrollment, setup, and token tests pass. The embedded scripts are exercised by the existing byte-identity and CA-contract tests. Compose smoke (now triggered for hub changes after #195) is the real-install gate; no live install was run.

---

> [!NOTE]
> **Medium Risk**
> Touches install-token expiry/cleanup, first-admin creation, and generated install commands; mistakes could break enrollment or lock out setup, but behavior is heavily tested and backward-compatible parsing is explicit.
> 
> **Overview**
> Hub onboarding reliability fixes: **token lifecycle**, **first-boot setup**, **PUBLIC_URL handling**, and **Linux install scripts**.
> 
> **Token expiry** — New mints store `expires_at` as UTC RFC3339 text via `tokenExpiryText` instead of passing `time.Time` through the driver. `parseTokenExpiry` and `checkTokenExpiry` accept legacy on-disk formats; periodic cleanup uses `deleteExpiredTokens` in Go rather than `expires_at < datetime('now')`, which mis-handled local-offset rows.
> 
> **First-boot setup** — `POST /api/setup` validates the token without consuming it first, consumes atomically only after password/PIN hashing, and **restores** the setup token if the admin insert fails so operators can retry. Concurrent setups still yield a single admin.
> 
> **PUBLIC_URL** — `publicAndWebsocketBase` strips trailing slashes so install/join URLs avoid `//install.sh`. Join drift checks compare normalized origins; legacy bindings with a trailing slash still serve scripts rebuilt from a normalized base.
> 
> **Linux bootstrap** — The dashboard paste block and embedded `install.sh` set `SUDO` from `id -u` and use `$SUDO` for privileged steps so **root** installs work in minimal containers without `sudo`.
> 
> Adds `hub/onboarding_repair_test.go` and updates installer tests for `$SUDO`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90ae42285e7e0520e5d8cb9a11b58de111bff41a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko (commented):
Approve at `d34ac0a` (COMMENT — leave merge to bokiko).

**Verdict:** the Unit B fixes look correct and well-tested.

- **Token expiry:** `tokenExpiryText` stores UTC RFC3339; `parseTokenExpiry` covers legacy driver-rendered offset/zone forms (including unnamed numeric zones); `deleteExpiredTokens` judges by instant so west-of-UTC fresh rows survive and east-of-UTC expired rows die. `checkTokenExpiry` shares the same parser. Unparseable rows left alone is the right call.
- **Setup retry:** validate/hash before consume; consume under `setupMu`; restore only when `setupTokenValue` is still empty after a failed insert. Concurrent test locks in single-admin. Validation/storage-failure survival test covers the retry path.
- **PUBLIC_URL:** single `TrimRight(..., "/")` in `publicAndWebsocketBase`; join drift compares trimmed mint binding so legacy slash spellings still serve; mint-time binding stays slash-free.
- **Root without sudo:** both paste block and embedded `install.sh` define `SUDO` from `id -u` before privileged steps; tests assert `$SUDO` and no bare `sudo`.

**Nits (non-blocking):** `deleteExpiredTokens` should check `rows.Err()` after the scan loop; prefer `t.Cleanup` to clear `setupStorageTestHook` so a failing test cannot leak the hook.

**Process:** stacked on `#196` (`fix/hub-lifecycle-safety`), which still has open request-changes. Retarget to `main` only after #196 is fixed/merged. Compose smoke / Hub race were still running at review time — confirm green before merge.

bokiko:
Final integration check after #196 squash merge: main tree1eb4d8a exactly matches the tested A head tree; git merge-tree(main,90ae422) produces faa58500, exactly the CI-tested B head tree. Actual main-to-result diff contains only the eight onboarding files, despite the stacked-history three-dot view listing A files. All exact-head checks including race and Compose passed. Legacy trailing-slash reconstruction fix independently checked. Merging this unit separately.

PR #200, opened 2026-09-07
agent: bound external discovery collectors, add agent race gate, release 6

bokiko:
Unit E, part 1 of the reliability plan (REPORT.md finding #13 plus the agent race CI gap). Agent and CI only; based on main.

## What changed

- **Bounded collectors** (`agent/command_plan.go`, `agent/main.go`, `agent/hardware.go`, `agent/hardware_windows.go`): every discovery command the metrics tick runs (nvidia-smi, docker info/ps, systemctl list-units, dmidecode ×3, lspci ×2, PowerShell CPU name) now goes through `runCollector`, a 10 s context timeout with the existing process-group kill and a `WaitDelay`. A hung collector previously stalled the connection loop, so pings stopped, the hub closed the socket after 90 s, and the agent never reconnected. On timeout callers report data unavailable, never zeroed readings; a hung nvidia-smi yields no GPU list.
- **Agent race gate** (`.github/workflows/ci.yml`): the Linux agent job runs `go test -race` like hub and proto.
- **Release 6** (`agent/release.go`): compiled agent bytes change, so the release and marker move from 5 to 6.

## Tests (`agent/collector_bounded_test.go`)

The test binary doubles as a synthetic collector (portable to the Windows runner): a hung child times out within the bound (0.3 s at a 300 ms test bound); a healthy child's output is returned; a hung nvidia-smi produces no GPU readings; the existing marker/constant check.

## Verification

`go vet` clean for linux, windows, and the insecure tag. In an isolated Go 1.26.8 Linux container: full agent suite with `-race` ok, insecure-tag suite ok, cross-compile of linux/amd64, linux/arm64, windows/amd64 with the release 6 marker present in all three. No live fleet or install actions.

## Not in this PR (Unit E part 2, own release number)

Windows `restart_service` stop-wait-start via the SCM API with an opt-in integration test on the windows-latest runner; bounded validation of the Linux rollback start-limit and the Windows pending-update helper; terminal user resolution fail-closed and supplementary groups.

---

> [!NOTE]
> **Medium Risk**
> Changes core connection-loop behavior and concurrent collector gating; mistakes could drop metrics or mis-handle timeouts, but scope is bounded subprocess discovery rather than auth or remote command execution (which already had separate timeouts).
> 
> **Overview**
> Adds **`runCollector` / `runBounded`** so metrics-tick subprocesses (nvidia-smi, docker, systemctl, dmidecode, lspci, Windows PowerShell) return within **10s**, kill on timeout, and refuse duplicate invocations while one is still outstanding (including unkillable D-state waits). Callers treat failures as **missing data**—e.g. hung `nvidia-smi` yields no GPU list instead of zeros—and the WebSocket loop can keep pinging instead of stalling until hub idle disconnect.
> 
> **CI** runs `go test -race` on the Linux agent job. **Release** bumps to **6** for the new binary. **`collector_bounded_test.go`** exercises timeouts, busy gating, fast-success vs race-with-deadline, and GPU hang behavior via a self-spawning test binary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bda013496668ac9874eff0391f609e15f8c431d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko (commented):
Approve at `e254e2b` (COMMENT — leave merge to bokiko).

**Verdict:** Unit E part 1 does what it claims.

- **`runCollector`:** 10s `CommandContext` + existing `configureCommand` process-group SIGKILL on Linux + `WaitDelay` 2s unblocks the connection loop even when a child ignores the kill (D-state `nvidia-smi` case). Callers return nil/unavailable, not zeros — covered by `TestGPUCollectorReportsUnavailableNotZeroOnHang`.
- **Call sites:** metrics-path discovery is converted (`nvidia-smi`, `systemctl`, `docker info/ps`, `dmidecode`×{17,sudo 17,16}, `lspci`×2, Windows PowerShell CPU). Remaining `exec.Command` usages (PTY bash, Windows updater helper, long-lived `power_nvidia` stream with its own stall restart) are out of this PR’s scope and already bounded differently.
- **Tests:** synthetic self-as-collector via `TestMain` + `BLOXOS_TEST_COLLECTOR` is portable; hang/healthy/GPU-unavailable cases look right.
- **CI:** Linux agent `go test -race` matches hub/proto. Release **5 → 6** marker/constant stay in sync.

**Nits (non-blocking):** `sendAll` still runs collectors on the connection `select` goroutine, so pings pause for the *sum* of per-collector bounds if several hang (~40s+WaitDelay, still under the hub’s ~90s idle). Acceptable for this fix; moving discovery off the ping loop would be a later hardening. Windows `configureCommand` remains a no-op — `CommandContext` + `WaitDelay` still apply.

CI was still running at review time — confirm Agent (incl. race) + Windows Agent green before merge.

bokiko:
Supervisor closure: the unkillable-child finding was fixed with an asynchronous bounded caller and single-outstanding per-command gate, not merely CommandContext/WaitDelay. Regression covers a worker that does not return at cancellation. A separate deterministic fast-success/cancel race was reproduced and fixed before merge. Exact merged head4bda013 passed agent race, hub race, Windows and Compose CI. Scope is external discovery collectors; in-process readers are not claimed bounded.

PR #201, opened 2026-09-07
fix: limit authenticated credential change attempts

bokiko:
Validation-queue follow-up, stacked on #199. Reproduced seven wrong current-PIN attempts receiving 403 with no limiting. Reuses existing limiter for five attempts per authenticated account per minute, separately for PIN and password; changing source IP does not evade it. Returns 429 and Retry-After with guidance. No onboarding steps or new infrastructure. Targeted race tests cover wrong attempts, correct request after exhaustion, source-IP changes, and existing successful password/PIN rotation. Retarget to main after #199; exact-head CI required before merge.

---

> [!NOTE]
> **Medium Risk**
> Touches auth credential-change handlers and abuse limits; behavior change could block legitimate rapid retries but reduces brute-force risk on authenticated endpoints.
> 
> **Overview**
> **Authenticated password and PIN change endpoints** now enforce a per-account attempt budget so stolen sessions cannot brute-force the current credential indefinitely.
> 
> Both `change-password` and `change-pin` call the existing `rateLimiter` with **five attempts per minute keyed by user ID** (not client IP), so rotating source addresses does not reset the budget. After the limit is hit, responses are **429** with **`Retry-After: 60`** and a short error message; wrong-current-credential paths still return **403** until the budget is exhausted.
> 
> A new integration test exercises both endpoints over seven attempts (including a would-be correct change on the last try), asserting **403** for early failures, **429** once limited, and that **`Retry-After`** is present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit addd66783e6c0fc3474557b50b212cefac368ef4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko (commented):
Approve at `3c99b24` (COMMENT — leave merge to bokiko).

**Verdict:** small, correct fix for unlimited wrong-current-credential probing on authenticated change-password / change-pin.

- Limit runs after JWT `userID` resolution and before bcrypt verify, so the budget is **per account** (not IP) and wrong attempts still pay bcrypt for the first five.
- Separate `change-password` / `change-pin` buckets via existing `rateLimiter.Allow(..., 5)` (1-minute window). `Retry-After: 60` matches that window.
- Test covers five 403s, then 429 (including a would-be-correct sixth/seventh), IP rotation, and `Retry-After`.

**Nits (non-blocking):** successful changes also consume the same budget (fine for this threat model). `Retry-After` is a fixed `60`, not derived from the oldest timestamp in the window (same pattern as other callers).

**Process:** stacked on `#199` → `#196`. Retarget to `main` only after those land. Confirm Hub tests green before merge.

PR #202, opened 2026-09-07
agent: SCM-driven Windows service restart, root without sudo, terminal identity (Unit E part 2)

bokiko:
Unit E part 2, stacked on #200 (release 7). Retarget to main after #200 merges.

**Windows service commands via the SCM.** `sc.exe stop; sc.exe start` raced the stop (stop returns while STOP_PENDING). Service commands now stop → poll STOPPED → start → poll RUNNING, bounded by the command context. `restart_service` on `BloxOSAgent` itself spawns a detached `-restart-service-helper` copy of the binary that performs the sequence from outside once the command response has flushed, logging to `<exe>.restart-helper.log`.

**Real SCM evidence in CI.** The windows job runs `TestSCM*` with `BLOXOS_SCM_TEST=1`: the test binary is installed as uniquely named services (slow 4 s stop, stuck stop, self-restarting instance, pending-update instance). Asserts restart waits for the slow stop and ends RUNNING under a new pid; a stuck stop is bounded with a state-naming error; stop/start wait for their states; the detached helper restarts a service from inside it; the update helper replaces a service binary under the SCM. Non-elevated with the gate set fails rather than skips.

**Windows pending-update helper.** Runner evidence: `timeout /t` returns at once without a console, so the old helper never waited. The helper now waits for SCM STOPPED, retries the move while the binary is locked, removes the marker only after the outcome is decided, starts the service and waits for RUNNING, logging every outcome to `<exe>.update-helper.log`.

**Root without sudo.** Plans built while running as root call `systemctl`/`docker`/`reboot` directly (the generated unit runs as root and root-onboarded hosts may have no sudo); unprivileged runs keep the sudo prefix. dmidecode tries sudo only when not root.

**Terminal identity (Linux).** No fallback to the agent's identity: `BLOXOS_TERMINAL_USER` must name an existing non-root account, otherwise the common non-root accounts are tried and the session is refused if none exists. Supplementary groups are set from the account's group ids; unparseable ids fail closed. Refusals go back as a `command_response` on the start_terminal id; the hub shows the reason in the browser terminal and closes the session instead of a 30 s blank wait.

**Linux rollback.** Reproduced on systemd 259 with disposable units on the smoke VM: a restart after start-limit-hit is refused until `reset-failed`. The served recovery script now runs `systemctl reset-failed` before `systemctl restart`. Signed release floors untouched.

---

> [!NOTE]
> **Medium Risk**
> Touches remote fleet control, Windows self-update/restart helpers, and Linux terminal identity; behavior changes are broad but heavily tested, including real SCM runs in CI.
> 
> **Overview**
> **Windows service commands** no longer use chained `sc.exe` argv plans (which raced `STOP_PENDING`). They go through the SCM with stop → wait **STOPPED** → start → wait **RUNNING**, bounded by the command timeout. Restarting the agent’s own service schedules a detached `-restart-service-helper` after the response flushes. CI runs opt-in `TestSCM*` against real services when `BLOXOS_SCM_TEST=1`.
> 
> The **pending-update batch helper** is rewritten for console-less runs: poll until **STOPPED**, retry the binary move while locked, then start and wait for **RUNNING**, with logging to `<exe>.update-helper.log`.
> 
> On **Linux**, remote plans drop the `sudo` prefix when the agent runs as root; **dmidecode** only retries via sudo when unprivileged. Self-restart/stop/reboot/shutdown no longer send a false `command_response` when the child dies by SIGTERM/SIGKILL during agent teardown—the hub can keep the truthful 202. **Terminals** never inherit root: `BLOXOS_TERMINAL_USER` must be a real non-root account (with supplementary groups), refusals are returned on the `start_terminal` id, and the **hub** surfaces that to the browser instead of a blank wait. Rollback script adds `systemctl reset-failed` before restart. Agent release bumps to **7**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92229117bafbadb72d38189d2dd2b5a9cecbdde9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko (commented):
Request changes at `63fcb28`.

**P2 — self-restart schedules the helper before the command response is written.** `platformServiceCommand` → `scheduleSelfRestart` → `spawnDetachedProcess` runs *before* `handleCommand` reaches `writeJSON`. The helper only sleeps `restartHelperDelay` (1.5s), while `writeJSON` can block up to `agentWriteWait` (10s) and may also wait on `writeMu` behind a metrics/ping write. Under backpressure the SCM can stop the agent mid-flush, so the hub never gets the success ACK the PR claims is sent “once the command response has flushed.”

Fix: write/flush the `command_response` first, *then* spawn the helper (or gate the helper on an explicit “response sent” signal / delay ≥ write wait). Add a regression that holds `writeMu` past 1.5s and asserts the helper has not been asked to stop yet (or that spawn happens only after unlock+write).

**Looks correct otherwise:**
- Windows service commands refuse argv (`errServiceViaSCM`) and go through SCM stop→wait STOPPED→start→wait RUNNING, bounded by ctx; stuck-stop error names `stop-pending`.
- Linux `commandPlanForIdentity` drops `sudo` when root; unprivileged keeps it. Tests cover both.
- Opt-in `BLOXOS_SCM_TEST=1` real-SCM suite (slow/stuck/self-restart/unknown) + fail-not-skip when gated but not elevated. Release **7** marker consistent.

**Scope note:** title mentions terminal identity; body correctly says that (plus Linux `reset-failed`, Windows pending-update evidence) is still to land — this review only covers SCM + root-without-sudo.

**Process:** stacked on `#200`. Retarget after #200 merges. Confirm Windows SCM CI step green.

Do not merge until the self-restart / response-flush ordering is fixed.

bokiko (commented):
Supervisor review at9222911: latest main including dashboard198 integrated; release7 marker retained. Independently reviewed SCM sequencing, helper STOPPED/locked-file/RUNNING flow, terminal credential/refusal interleavings and the new structured signal-death discriminator. Real Windows SCM tests execute (not skipped) and pass. On isolated Linux I independently ran the final signal/ordinary-error/other-service/timeout regressions and a real PTY root-to-user test: uid501, intended groups1000/999 and home all correct. Full combined hub race passed488.895s; optimized dashboard browser suite passes. Candidate serves matching exact-hash release7 Linux amd64/arm64 and Windows amd64 binaries. The earlier Compose failure remains a hard gate: require unchanged self-restart202/new-PID smoke to pass on this head, plus all remaining CI, before merge. No live fleet rollout.

bokiko (commented):
Request changes at `9222911` (COMMENT — leave merge to bokiko).

**Prior P2 still open: self-restart helper is still spawned before `command_response` is written.**

Order in `handleCommand`:
1. `platformServiceCommand` → `scheduleSelfRestart()` → `spawnDetachedProcess(...)` (helper running)
2. then `writeJSON(conn, mu, resp)`

The helper only `time.Sleep(restartHelperDelay)` with `restartHelperDelay = 1.5s`, while `agentWriteWait = 10s` and `writeMu` can already be held by metrics/`sendAll`. If the response write is delayed past 1.5s, the helper stops the service mid-write and the hub can lose the ACK (or see a drop with no confirmation).

PR/docs say the helper runs “after the command response has flushed,” but the code still uses spawn-then-hope-the-delay-wins — same race as the prior review. No commit on this head reorders spawn after a successful write.

**Fix shape (either works):**
- Prefer: return “restart scheduled” without spawning; `writeJSON` first; spawn helper only after the write succeeds (or on a `defer` after write returns).
- Or: make the helper wait on a real signal (named event / parent exit + write complete), not a fixed 1.5s that is shorter than `agentWriteWait`.

**What does look good on this head (non-blocking):**
- SCM stop→STOPPED→start→RUNNING for non-self targets; stuck-stop bounded with state-named errors.
- Update helper: wait STOPPED, retry locked move, then RUNNING + log — real SCM tests in CI.
- Linux: root without sudo; terminal identity fail-closed + refusal as `command_response` on start_terminal id; hub surfaces it.
- `shouldSuppressTeardownReply` for SIGTERM/SIGKILL self-teardown (Linux) — keeps hub 202 truthful.

Windows Agent Tests / Agent / Hub / Compose smoke green; Hub race still finishing at review time. Do not merge until the flush ordering is fixed.

PR #214, opened 2026-09-09
fix: recover onboarding trust and prevent incompatible agent installs

bokiko:
## Summary

Fix the product paths behind two onboarding failures without deleting machine trust or weakening enrollment authentication.

- Linux bootstrap preserves a mismatching legacy CA and installs the newly fingerprint-verified CA at a separate content-addressed path. The installer persists that path in the service; secrets, update keys, and rollback floors stay intact.
- Native/Compose Caddyfiles retain RSA leaf keys across ordinary renewal, so an otherwise fresh 15-minute pinned command does not break at renewal. Native IP clients also receive the configured default SNI certificate. Explicit key rotation still fails closed and needs a fresh command.
- Both installers request `enrollment=1` when downloading the agent. The hub refuses an unnumbered binary on that path because it cannot establish enrollment-handshake compatibility. The existing flagless signed-update/migration download stays unchanged.
- Hub logs distinguish incomplete enrollment from ordinary disconnection, with possible causes and recovery instructions. Dashboard/docs explain stale pins, inconsistent TLS endpoints, retained CAs and native payload activation.

## Why

Older deployed agents save an issued secret but never send `enrollment_committed`. A newer hub correctly refuses to commit that enrollment, then closes at its 30-second authentication deadline. The agent keeps running but cannot authenticate subsequent reconnects. Serving that old payload again repeats the problem. This change prevents the incompatible installation rather than undoing the durable enrollment handshake.

Not every unnumbered build is incompatible; this is a conservative compatibility check. Numbered official builds postdate the commit handshake. It is not an authentication or signing substitute.

## Validation

- Exact-commit CI: hub tests, full hub race tests, both agent platforms, shared protocol, dashboard and Compose onboarding all passed.
- Executed generated Bash recovery tests: new CA, matching legacy CA, rotated CA, cached replacement, bad download, corrupt cached target, and preservation of credential/floor files.
- Real curl test: renewed certificate with the same key works; changed key cannot execute the script.
- Final focused race run passed (14.9s), covering compatibility gates, both changed URL expectations, CA recovery, pin renewal and bundled Caddy configuration.
- An earlier local full race run started during edits and failed only the two old installer-URL expectations. The committed expectations are corrected; no race warning was reported. Full race CI passed on the exact committed snapshot.
- Dashboard: 94 tests, TypeScript, component ESLint and production build passed.
- Both Caddyfiles adapted with verified official Caddy 2.11.4; generated policies retain RSA2048 and `reuse_private_keys`.
- Unchanged agent source cross-builds for Linux/Windows; native macOS agent tests are unsupported. Linux CI covers the agent runtime tests.

## Deployment / limits

- No live machine, CA, credential, service, update floor or fleet signing key was changed.
- Existing native hubs must activate compatible official agent payloads through their normal staging/signing procedure; updating only the hub executable does not replace those files.
- Already-stranded machines need a fresh Add Machine command after payload activation. The patch cannot recover a secret the old handshake never committed.
- Apply/reload the Caddyfile separately on existing deployments. Preserve Caddy's CA data. `reuse_private_keys` trades per-renewal leaf-key rotation for onboarding stability and is documented by Caddy as temporary; review it when upgrading Caddy.
- Windows CA rotation remains operator-managed; the automatic retained-CA recovery here is Linux-only.

### DIFF
--- a/dashboard/src/components/AddMachineModal.tsx
+++ b/dashboard/src/components/AddMachineModal.tsx
@@ -591,7 +591,8 @@ function CommandDisplay({
                 TLS key (<code className="font-mono text-[10px] break-all">{resp.join_pin}</code>), so a different
                 server can&apos;t answer in its place. Hub certificates renew periodically; if the machine prints{" "}
                 <code className="font-mono text-[10px]">curl: (90) SSL: public key does not match pinned public key</code>
-                , the key changed since this command was made. Come back here and generate a new one.
+                , the endpoint no longer matches this command. Generate a new one. If a fresh command also fails,
+                ask the hub administrator to check the proxy certificate and pin-dial address. Do not remove the pin.
               </p>
             )}
             <p>
--- a/docker/Caddyfile
+++ b/docker/Caddyfile
@@ -13,6 +13,9 @@ https://{$HUB_HOST} {
 	# accept Caddy's default ECDSA keys (see AGENTS.md).
 	tls internal {
 		key_type rsa2048
+		# Join commands pin the leaf key for 15 minutes. Keep that key
+		# across routine renewals so a renewal cannot invalidate a fresh command.
+		reuse_private_keys
 	}
 
 	# Hub API, agent and terminal WebSockets, installers and downloads.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,6 +58,43 @@ treats the hub as public. If the hub cannot reach `PUBLIC_URL` over TLS at all,
 it refuses to mint rather than guess — this is a connectivity problem to fix,
 not a trust decision.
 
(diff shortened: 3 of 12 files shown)


bokiko:
Engineering review: Claude checked the committed CA/pin recovery diff and found no blockers; its 5 compatibility tests and enrollment/handshake sweep pass. Codex checked the gate changes and corrected wording to distinguish unverified compatibility from known incompatibility. Local Qwen review was rechecked against the complete bootstrap trust context; no substantiated new blocker remained. Kimi was requested but hit its quota before completing a review. Both Caddyfiles include default_sni (Compose already had it). Agent source/release number unchanged; no fleet deployment performed. Waiting for exact-commit CI, including Compose and hub race, before merge.

PR #216, opened 2026-09-09
feat: verified upgrade foundation (public build and instance checks)

bokiko:
## Phase 1: identify the installation actually serving the website

Phase-1 foundation for the agreed verified-upgrade workflow. **Not a complete native/Compose updater; merging this does not mean legacy upgrades are solved.**

### Problem

Updating Compose images can succeed while native hub/dashboard services continue serving the public URL. Container health and checkout tags do not establish which build the browser reaches. Existing `hub_sha` fields describe agent payloads, not the hub executable.

### Changes

- Public, uncached hub `/api/build-info` and dashboard `/build-info`: component, build version/revision, process instance ID.
- Build stamps wired into release images without changing agent compilation flags; dashboard stamps compiled in rather than read from runtime environment.
- Read-only public/direct identity verifier, verified TLS and private-CA support, no redirect/proxy ambiguity, unknown legacy metadata never passes.
- Regression fixtures for wrong backend, same build/different instance, malformed metadata, CA verification and failure precedence.
- Production standalone dashboard test covers runtime override and restart identity; Compose release smoke compares public identities to selected containers.
- README explicitly distinguishes Compose updates from native installations and no longer equates healthy containers with a verified website upgrade.

### Local validation

- Dashboard: 97 unit tests pass; production build passes; lint has no errors (three existing unused-disable warnings).
- Standalone runtime test passes, including deliberately misleading runtime version/revision and different instance IDs after restart.
- Hub `go vet ./...` passes; focused `TestBuildInfo` tests pass under the race detector.
- Real HTTPS integration checks pass using isolated local servers.
- All 24 upgrade verifier tests pass (20 regression tests plus 4 real HTTPS tests). The added real truncated chunked-response test reproduced the crash before the fix and passes after it.
- Workflow YAML parses, shell syntax and diff whitespace checks pass.
- All seven test/build checks passed on the final interrupted-response fix, including full hub race and disposable Linux Compose smoke.

### Collaboration

Claude implemented/reviewed hub identity and reviewed dashboard/release/docs changes. Kimi implemented the verifier and regression fixtures; Codex reviewed and required corrections for ignored CA context, public-only false success, unknown metadata, URL validation and failure precedence. Existing visible Herdr sessions were reused. No production host was changed.

Both reviewers approved the final interrupted-response handling and real regression test after their full review. User authorized merge once we agree and checks pass. Native/Compose update adapters remain separate follow-up work.

### Still required before the full upgrade workflow can ship

- Supported native/Compose target discovery and ownership evidence, including the first hop from releases without identity endpoints.
- Separate in-place update adapters behind a simple entry point; staging both components before downtime.
- Correct native/Compose backup boundaries and migration-aware rollback/interruption testing.
- Mixed-host end-to-end update regression (not just metadata fixtures), agent payload/signing status reporting, complete release validation.

No automatic migration, proxy repointing, orphan cleanup, key deletion or fleet rollout is included.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,9 @@ jobs:
       - name: Test backup failure and restart handling
         run: python3 ../scripts/test_backup_compose.py
 
+      - name: Test wrong-installation upgrade safeguards
+        run: python3 -m unittest discover -s ../scripts/tests -p 'test_upgrade*.py'
+
       - name: Test native agent bundle validation and staging
         run: |
           python3 ../scripts/test_agent_bundle.py
@@ -202,4 +205,13 @@ jobs:
         run: pnpm test
 
       - name: Build dashboard
+        env:
+          BLOXOS_BUILD_VERSION: ${{ github.ref_name }}
+          BLOXOS_BUILD_REVISION: ${{ github.sha }}
         run: pnpm build
+
+      - name: Verify production dashboard build identity
+        env:
+          EXPECT_VERSION: ${{ github.ref_name }}
+          EXPECT_REVISION: ${{ github.sha }}
+        run: node scripts/test-build-info.mjs
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -55,6 +55,9 @@ jobs:
         with:
           context: .
           file: Dockerfile.hub
+          build-args: |
+            BLOXOS_BUILD_VERSION=${{ github.ref_name }}
+            BLOXOS_BUILD_REVISION=${{ github.sha }}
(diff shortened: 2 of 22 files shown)


bokiko:
Final review reconciliation: Claude and Kimi both approved the interrupted-response fix and real chunked-response regression at b9ef60f3701efd665f92e7df3b3f26f34053869b. Codex reproduced the original failure before applying the handler; all 24 local verifier tests now pass. This is phase-1 read-only verification, not the complete updater. User authorized merge after agreement; waiting for the full checks on this exact head before merging. No live deployment or fleet changes.

PR #218, opened 2026-09-09
fix: package native servers by platform child digest

bokiko:
Fixes the v1.3.0 native packaging failure on classic Docker stores: pulling both architectures under the same OCI index digest fails with cannot overwrite digest. Resolve exactly one supported child manifest per architecture, pull/inspect/create by that immutable child, and retain canonical index digests in the update manifest. Reject ambiguous, missing, malformed or mismatched platforms. Four regression tests added; all 134 updater/preflight/TLS tests pass. No runtime updater changes. Actual v1.3.0 image extraction and Claude review are in progress. The original release tag and image bytes remain unchanged.

### DIFF
--- a/scripts/export-server-bundle.py
+++ b/scripts/export-server-bundle.py
@@ -16,8 +16,28 @@ def run(args):
     return subprocess.check_output(args, text=True).strip()
 
 
+def platform_image(image, platform):
+    """Resolve the immutable index to one child, avoiding classic-store collisions."""
+    index = json.loads(run(["docker", "manifest", "inspect", image]))
+    os_name, arch = platform.split("/")
+    matches = []
+    for item in index.get("manifests", []):
+        target = item.get("platform", {})
+        if target.get("os") == os_name and target.get("architecture") == arch:
+            if target.get("variant", "") not in (("", "v8") if arch == "arm64" else ("",)):
+                continue
+            matches.append(item.get("digest", ""))
+    if len(matches) != 1 or not re.fullmatch(r"sha256:[0-9a-f]{64}", matches[0]):
+        raise ValueError("Published image index must contain exactly one supported " + platform + " manifest")
+    return image.split("@", 1)[0] + "@" + matches[0]
+
+
 def export(image, platform, source, destination, revision):
+    image = platform_image(image, platform)
     subprocess.run(["docker", "pull", "--platform", platform, image], check=True)
+    actual_platform = run(["docker", "image", "inspect", "--format", '{{.Os}}/{{.Architecture}}', image])
+    if actual_platform != platform:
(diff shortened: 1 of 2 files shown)


bokiko:
Actual extraction of both amd64 and arm64 hub/dashboard bundles from the immutable v1.3.0 image indexes now passed in the disposable Linux VM. All 134 local tests passed. Claude independently reviewed the exact fix and approved it with no blockers. This is packaging-only: image bytes, release tag/revision, update manifest index references and runtime updater remain unchanged. The complete asset set is regenerated consistently, not mixed with any partial failed output.

bokiko (commented):
## Engineering review @ `d66a4b4`

**Approve** (COMMENT). Do not merge from this lane.

Tight packaging fix for classic Docker content-store collisions when pulling both arches under one OCI index digest.

### Verified
- `platform_image` resolves the index to **exactly one** supported child (`linux/amd64` empty variant; `linux/arm64` empty or `v8`), ignores attestation `unknown/unknown`, and fails closed on missing / duplicate / bad digest / unexpected variant (`v9`).
- `export` pulls, platform-checks, revises, and `create`s against the **child** digest; update-manifest still records the caller’s **canonical index** digests (`args.hub` / `args.dashboard`) — correct for release identity.
- Callers already require `ghcr.io/bokiko/bloxos-{hub,dashboard}@sha256:…`; only `linux/amd64` and `linux/arm64` are packaged.
- Four new unit tests cover distinct children, fail-closed cases, child used for pull/create, and wrong-platform rejection before create.
- No runtime updater path changes.

### CI at review time
Hub / Agent / Windows Agent / Protocol / Dashboard green; Compose smoke + hub-race still running (docs/scripts-only risk is low — wait for green before merge).

### Nits
1. PR body notes live **v1.3.0 image extraction still in progress** — worth confirming one amd64+arm64 export on a classic store before cutting the release asset, even though the unit coverage is solid.
2. Single-arch (non-index) manifests fail closed via empty `manifests` — fine for current digest-pinned multi-arch inputs; just don’t feed a child digest as `--hub`/`--dashboard`.

Ship when CI is green and you’re happy with a real dual-arch extract.

bokiko:
Release recovery verified: all 9 v1.3.0 assets downloaded back from GitHub passed checksum/architecture/source-linkage verification. The documented public latest/download/bloxos-update bootstrap then updated the disposable native Linux VM successfully. Both public hub and dashboard report v1.3.0 at 8d110a8c7d05a821b84035f3fbe1849049f1f5d9, preserving the DB seed, ownership and CA. v1.3.0 is published with an explicit packaging-recovery note. No production deployment changed. Waiting for this PR’s remaining race check before merge.

PR #219, opened 2026-09-09
fix: prove a bare-IP public edge by address ownership

bokiko:
## The failure

Any install whose `PUBLIC_URL` is a bare IP — the common LAN case — could not complete `init`, and every preflight failed:

```
BloxOS updater: cannot prove the public edge still routes to this host's hub:
could not compare local and public certificates: TLS probe of 192.168.16.113 failed: SSLError
```

Reported from a real v1.3.0 upgrade on an IP-based deployment.

## Root cause

SNI may not carry an IP literal (RFC 6066), so CPython sends **no SNI at all** for one. Verified empirically:

```
server_hostname='127.0.0.1'  -> SNI received by server: None
server_hostname='localhost'  -> SNI received by server: 'localhost'
```

With no SNI, the proxy selects its certificate by the **local address the connection landed on**. Three observations from the affected host pin this down exactly:

| connection | SNI | result |
|---|---|---|
| `curl` → `192.168.16.113:443` | none | 200 |
| `openssl` → `127.0.0.1:443`, `-servername 192.168.16.113` | IP | established |
| `openssl` → `127.0.0.1:443`, `-noservername` | none | internal error |

So dialling the public IP serves correctly, but the **loopback leg** of the local-vs-public comparison finds no certificate for `127.0.0.1` and the handshake is refused:

```python
local  = leaf_cert("127.0.0.1", 443, host, ca_file)   # dies for an IP edge
remote = leaf_cert(host, port, host, ca_file)          # would have succeeded
```

The comparison is simply **not expressible** for an IP edge.

## The fix

For a bare-IP public URL, prove the edge **structurally** rather than by certificate — a *stronger* claim than certificate equality:

- the public address must be assigned to this host (an ephemeral `bind()` succeeds only for a locally configured address — no `ip addr` parsing, no new dependency);
- combined with the already-proven facts (the verified-owned proxy holds `:443`, and its **active** config routes to our owned loopback backends), an IP bound here cannot be served anywhere else.

Two hosts can share a certificate; one LAN address binds in one place.

DNS hostnames are **unaffected** and keep the certificate comparison, where SNI works and the check genuinely detects a name pointing at another machine. A public IP that is *not* local is refused with an actionable message instead of an opaque `SSLError`.

## Tests

5 new cases: IPv4/IPv6/non-default-port IP edges verify without ever attempting a TLS probe; a non-local public IP is refused; a DNS edge still performs both certificate probes; plus `_is_ip_literal` and the real `_address_is_local` (loopback true, RFC 5737 TEST-NET-1 false).

Full updater suite: **138 passed**.

---

> [!NOTE]
> **Medium Risk**
> Changes init/preflight public-edge proof logic on the security-sensitive updater path, though DNS verification is unchanged and the new IP path replaces a previously broken check.
> 
> **Overview**
> Fixes **init and preflight** for LAN installs where `PUBLIC_URL` is a bare IP (IPv4/IPv6), which previously failed with opaque TLS errors because the local-vs-public certificate check cannot work when Python sends no SNI for IP literals.
> 
> **`_verify_routing`** now branches on **`_is_ip_literal`**: for DNS hostnames, behavior is unchanged (compare local `:443` and public leaf certs). For a bare IP, it **skips TLS probes** and instead requires **`_address_is_local`**—an ephemeral `bind()` that succeeds only if the address is configured on this host—together with the existing Caddy upstream checks.
> 
> **`discover_native`** exposes an injectable **`address_is_local`** seam (default **`_address_is_local`**) for tests. New unit tests cover IP edges (with/without custom port, IPv6), refusal when the IP is not local, unchanged DNS cert probing, and helper behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73b53c540c70155ca6ba46c06cf8b36b91ee80b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/tests/test_updater_native.py
+++ b/scripts/tests/test_updater_native.py
@@ -439,6 +439,60 @@ class DiscoverTests(unittest.TestCase):
         self.assertFalse(valid)
         self.assertNotEqual(native._norm_hostport("[::1]:4000"), native._norm_hostport("127.0.0.1:4000"))
 
+    # --- bare-IP public edge (LAN installs) --------------------------------
+    # SNI may not carry an IP literal (RFC 6066), so Python sends NO SNI for one
+    # and a Caddy site keyed by an IP kills the handshake. A bare-IP edge must
+    # therefore be proven by local address ownership, never by a TLS probe.
+    LOOPBACK_UPSTREAMS = {"upstreams": [{"dial": "127.0.0.1:4000"}, {"dial": "127.0.0.1:3000"}]}
+
+    def _route_ip(self, public_url, is_local):
+        probed = []
+
+        def leaf_cert(*args):
+            probed.append(args)
+            raise AssertionError("a bare-IP edge must never be probed over TLS")
+
+        valid, reason = native._verify_routing(
+            public_url, "http://127.0.0.1:4000", "http://127.0.0.1:3000", None,
+            lambda: self.LOOPBACK_UPSTREAMS, leaf_cert, lambda host: is_local)
+        self.assertEqual(probed, [])
+        return valid, reason
+
+    def test_bare_ip_edge_is_proven_by_local_address_not_tls(self):
+        for url in ("https://192.168.16.113", "https://192.168.16.113:8443", "https://[fd00::1]"):
+            with self.subTest(url=url):
+                valid, reason = self._route_ip(url, True)
+                self.assertTrue(valid, reason)
+                self.assertIn("bound to this host", reason)
+
+    def test_bare_ip_edge_not_on_this_host_is_refused(self):
+        valid, reason = self._route_ip("https://192.168.16.113", False)
+        self.assertFalse(valid)
+        self.assertIn("not configured on this host", reason)
+
+    def test_dns_edge_still_compared_by_certificate(self):
+        probed = []
+
+        def leaf_cert(host, port, server_name, ca_file):
+            probed.append((host, server_name))
+            return b"same verified certificate"
+
+        valid, _ = native._verify_routing(
+            "https://hub.example", "http://127.0.0.1:4000", "http://127.0.0.1:3000", None,
+            lambda: self.LOOPBACK_UPSTREAMS, leaf_cert,
+            lambda host: self.fail("a DNS edge must be proven by certificate"))
+        self.assertTrue(valid)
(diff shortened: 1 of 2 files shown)


bokiko (commented):
## Engineering review @ `73b53c5`

**Approve** (COMMENT). Do not merge from this lane.

Root cause matches the failure mode: CPython sends no SNI for IP literals (RFC 6066), so the loopback leg of local-vs-public leaf comparison is not expressible for bare-IP `PUBLIC_URL`. Branching on `_is_ip_literal` is the right split.

### Verified
- DNS path unchanged: still compares `leaf_cert(127.0.0.1:443, SNI=host)` vs `leaf_cert(host:port, SNI=host)`; tests assert both probes and that `address_is_local` is not consulted.
- IP path: still requires active Caddy upstreams → owned loopback hub/dashboard; then `_address_is_local` via ephemeral `bind()` (EADDRNOTAVAIL fail-closed). No TLS probes (tests assert).
- Non-local public IP refused with an actionable message (no more opaque `SSLError`).
- `_is_ip_literal` treats bracketed IPv6; `localhost` / DNS stay on the cert path.
- Injectable `address_is_local` seam on `discover_native` / `_verify_routing`; real loopback vs RFC 5737 TEST-NET-1 covered.
- Discover still requires `:443` (and admin `:2019`) in the selected proxy unit’s cgroup before routing proof — so the IP branch is not “bind alone.”

### CI at review time
Agent / Windows Agent / Protocol / Dashboard green; Hub / hub-race / Compose still running (scripts-only — wait for green).

### Nits (non-blocking)
1. **Custom port on IP edges is unused.** `port = pub.port or 443` is ignored in the IP branch, so `https://192.168.x.x:8443` can verify while all listener ownership is about `:443`. Either document that IP edges are :443-only, or include `port` in the IP proof.
2. **Optional hardening:** after `address_is_local`, a single TLS dial to the public IP:port (no loopback compare — that dial works without SNI) would catch “IP is assigned but nothing answers on the edge.” Not required for this fix; the previous behavior was total failure.

Ship when CI is green. Good LAN unblock for v1.3.0 IP installs.

PR #223, opened 2026-09-09
fix(hub): refuse an Add Machine command the hub cannot fulfil

bokiko:
## The failure, from the field

A hub serving a markerless agent minted a join command happily. The operator ran it on two real machines and only there learned the hub refuses the download:

```
Agent download failed (HTTP 503) for arch=amd64
this hub's linux/amd64 agent binary (source environment:BLOXOS_AGENT_BINARY)
carries no release marker, so the hub cannot verify it supports the enrollment
handshake...
```

The download gate is right to refuse — a binary too old to complete the enrollment handshake enrols, looks `active (running)`, then drops offline ~30s later and fails every reconnect. But nothing said so until the install was already underway on the target machine.

## The fix

`handleCreateToken` already mints nothing when it cannot produce a trustworthy command — no `PUBLIC_URL`, an unclassifiable CA — on the stated principle that *"a hub that cannot produce a trustworthy command mints nothing"*. Agent availability belongs in that list.

The mint path now applies exactly the condition `handleDownloadAgent` enforces, and refuses with the remedy: which platform, which source, the staging document, and the reassurance that already-enrolled agents keep updating.

## Deliberately narrow

Only a **markerless** served binary blocks minting:

- An architecture that resolves to **nothing at all** is a different, pre-existing condition — the download answers it with a 404 naming every path it looked at. Widening the refusal to cover that would block Add Machine on hubs that work today, including every hub with no arm64 payload staged.
- **One servable architecture is enough.** An amd64-only fleet is legitimately unaffected by a missing or markerless arm64 payload; the other platforms are logged instead.

## Testing

The resolver requires a root-owned binary — correctly — so a unit test cannot produce a resolvable fixture. Rather than fake root or install a shared package-level seam that parallel tests could race on, the classification takes its state lookup **as a parameter** and the refusal is a **pure decision** over the result. Both are tested directly:

- markerless is blocking (and names its source so an operator can find it)
- release-marked is servable
- unresolvable is neither
- Windows never affects a Linux join command
- refuse only when nothing is servable *and* something is markerless
- the message is actionable on its own
- **non-regression:** the default environment resolves no binary and still mints exactly as before

`go vet` clean; full hub suite passes.

---

> [!NOTE]
> **Medium Risk**
> Changes Add Machine token minting and API response shape; behavior is fail-closed only when served binaries are markerless, with explicit non-regression for unresolved binaries.
> 
> **Overview**
> **Add Machine** now checks agent binary readiness **before** minting a token, using the same enrollment rules as agent download—so operators are not given install commands that will fail with HTTP 503 on the target.
> 
> If every **served** platform binary lacks a release marker (enrollment cannot be verified), token creation returns **503** with an actionable error (platform, source, staging docs). Missing binaries alone still allow minting (unchanged). If at least one platform is enrollment-ready, minting proceeds; partially broken platforms are listed in a new **`unusable`** field on the install token response and logged.
> 
> Classification is factored into testable helpers with injected binary state; unit and integration tests cover refusal, partial success, and the no-binary-resolves non-regression case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 545a89430adaaa31c53122ace1da99a47303ded2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/hub/agentws.go
+++ b/hub/agentws.go
@@ -157,6 +157,22 @@ func (s *Server) handleCreateToken(c echo.Context) error {
 		})
 	}
 
+	// A command that cannot deliver an installable agent is not a trustworthy
+	// command either: the operator runs it on a real machine and only there
+	// discovers the hub refuses the download. A markerless binary can never
+	// prove it supports the enrollment handshake, so when no Linux
+	// architecture can be served, mint nothing and say exactly how to fix it.
+	servable, markerless := enrollmentAgentReadiness()
+	if reason := enrollmentRefusal(servable, markerless); reason != "" {
+		log.Printf("token_create: refusing to mint, no Linux agent can prove enrollment support: %s",
+			strings.Join(markerless, "; "))
+		return c.JSON(http.StatusServiceUnavailable, map[string]string{"error": reason})
+	}
+	if len(markerless) > 0 {
+		log.Printf("token_create: minting; these platforms cannot serve a fresh install: %s",
+			strings.Join(markerless, "; "))
+	}
+
 	// Everything the join command needs is settled before the token row
 	// exists, so a hub that cannot produce a trustworthy command mints
 	// nothing: no token, no link, no fallback to an unauthenticated fetch.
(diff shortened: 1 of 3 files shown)


bokiko (commented):
## Review — approve

Right place to fail closed. Minting a join command the download path will refuse with `?enrollment=1` + markerless binary is exactly the “hub that cannot produce a trustworthy command mints nothing” rule already used for `PUBLIC_URL` / CA classification. Matching `Release < minEnrollmentAgentRelease` against `handleDownloadAgent` is the correct shared condition.

### What looks solid
- **Narrow scope:** only Linux, only markerless (resolved path+SHA+source but release 0). Missing/unresolvable arch stays a download 404 so amd64-only fleets keep minting.
- **One servable arch is enough** — partial markerless is log-only; refusal needs `len(servable)==0 && len(markerless)>0`.
- **Fail before token insert** — readiness runs before CA classify / INSERT (same ordering discipline as the TLS gate).
- **Operator message** names platform, source, staging doc, and that enrolled agents keep updating.
- **Testability:** injected `enrollmentReadinessFrom` + pure `enrollmentRefusal` avoid root-owned fixtures; classification, refusal matrix, Windows-ignored, and non-regression mint when nothing resolves are all covered.

### P2 — Windows Add Machine is collateral
`handleCreateToken` returns **both** Linux and `WindowsCommand`. Classification only looks at Linux, so a hub with **markerless Linux + fine Windows** refuses the whole mint — including a Windows install that downloads `/download/agent?os=windows` **without** `enrollment=1` and would not hit this gate today.

Acceptable for the field failure (Linux stranding), but worth a follow-up: document that Add Machine now requires an enrollment-ready Linux agent, or make the response OS-aware so a Windows-only install is not blocked by a Linux-only staging problem. Not a reason to hold this PR.

### Nits
- Refusal copy uses singular “carries” after a joined list of platforms.
- No end-to-end HTTP 503 integration with an injected markerless state (unit coverage of the decision is enough given resolver root requirements).

**Verdict: approve.** COMMENT only — not merging.

PR #224, opened 2026-09-10
fix(dashboard): let Ledger's absent readings read as absent

bokiko:
Two problems that were only visible once the layout was actually rendered in a browser.

**1. `N/A` carried the weight of a measurement.** An unavailable reading set at the full 1.9rem figure size, so two dead fields (GPU power, AI sessions) competed with live data and the eye stopped on nothing. It now sets smaller and muted, holding its place in the row without pretending to be a number.

**2. A fully stale fleet looked broken.** Averages exclude stale machines by design, so when nothing is fresh every figure reads `N/A` while the table below still lists per-machine values. That is correct and it looks like a bug. The band now explains itself in place:

> No machine has reported inside the fresh window, so no fleet average can be computed. The per-machine readings below are the last values received, not current ones.

The closing footnote already stated the rule; this states the moment.

Verified in a browser against demo data in both states — fresh (figures live, N/A recedes) and fully stale (note present, statuses `Stale`, every average `N/A`) — rather than trusting a green build. Dashboard 106 tests, lint clean, `next build` succeeds.

---

> [!NOTE]
> **Low Risk**
> Ledger-only presentation and copy changes; no aggregation logic or API behavior is modified.
> 
> **Overview**
> Improves how the **Ledger** fleet layout communicates missing data so operators are not misled by correct but confusing UI.
> 
> **N/A** fleet figures (CPU averages, GPU power, etc.) now get the `ll-absent` treatment—smaller, muted type—so unavailable readings keep their slot in the row without competing visually with live measurements. The shared `Figure` helper applies that class when `value === "N/A"`.
> 
> When every enrolled machine is stale (nothing inside the fresh window), a **`ll-stale-note`** appears above the "Running now" figures explaining that fleet averages cannot be computed and that the table still shows last-received per-machine values. Styling for both behaviors is scoped under `[data-layout="ledger"]` in `design-layouts.css`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 121b8338ecc385e0e4d33b07581c63fd3d8d6187. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dashboard/src/app/design-layouts.css
+++ b/dashboard/src/app/design-layouts.css
@@ -501,6 +501,19 @@ html[data-layout="ledger"] body { background: var(--ll-canvas); color: var(--ll-
 [data-layout="ledger"] .ll-unit {
   margin-left: 0.2rem; font-size: 0.85rem; font-weight: 400; color: var(--ll-secondary);
 }
+/* An unavailable reading keeps its place in the row but not its weight, so
+ * the eye passes over absence and stops on measurements. */
+[data-layout="ledger"] .ll-figure strong.ll-absent {
+  font-size: 1.05rem; color: var(--ll-muted); letter-spacing: 0;
+  line-height: 1.75; /* hold the baseline of a full-size figure */
+}
+/* Averages exclude stale machines, so when nothing is fresh every figure is
+ * N/A while the table below still shows per-machine numbers. Correct, but it
+ * reads as broken without a reason given in place. */
+[data-layout="ledger"] .ll-stale-note {
+  margin: 0 0 1.5rem; max-width: 68ch; font-size: 0.82rem; line-height: 1.55;
+  color: var(--ll-secondary); border-left: 2px solid var(--ll-rule-strong); padding-left: 0.85rem;
+}
 [data-layout="ledger"] .ll-figure p {
   margin: 0; font-size: 0.78rem; line-height: 1.45; color: var(--ll-muted);
 }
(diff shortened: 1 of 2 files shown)


bokiko (commented):
## Review — needs-work

The presentation intent is right: unavailable figures should not compete with live ones, and a fleet with no fresh samples needs an in-band explanation so the table does not look like a bug. `ll-absent` + the stale note are the right shape, scoped under `[data-layout="ledger"]` only.

### What looks solid
- `Figure` marks `value === "N/A"` (what `num(null)` and the sessions null path already emit); live Connected counts stay full weight.
- `ll-absent` keeps row geometry (`line-height: 1.75`) while dropping size/colour — matches the “hold the place, lose the weight” goal.
- Stale-note condition `agg.online - agg.stale <= 0` is correct for a **fully stale connected** fleet (`stale ⊆ online` ⇒ `online === stale`).
- Dashboard Lint + Build already green on this head. No aggregation/API changes.

### P2 — all-offline contradicts Connected copy
When every machine is **offline** (`online === 0`, `stale === 0`, `total > 0`):
- the new `ll-stale-note` correctly appears (“no fleet average…”)
- Connected’s note still falls through to **`"All reporting fresh"`** because it only special-cases `stale > 0`

That is exactly the class of misleading adjacent copy this PR is fixing. Pre-existing Connected note, but this PR puts the honest band note next to it.

Fix before merge (either is fine):
- Connected note: when `agg.online === 0`, say something like “None connected” / “No fresh reports”, or
- tighten the band note to `agg.stale > 0 && agg.online === agg.stale` if you only want the fully-stale case from the PR body.

Nit: no unit test for the note predicate; optional for a pure presentation change.

**Verdict: needs-work** (Connected/offline honesty). COMMENT only — not merging.

PR #226, opened 2026-09-10
Replace five layouts with the Monoform design system

bokiko:
Replaces BloxOS's five selectable layouts (Classic, Operations Wall, Grove Workspace, Precision Console, Ledger) and its palette gallery with **one** visual system: **Monoform**, in two contrast modes that share identical components, geometry and interaction — `gray` (default) and `dark`. No light mode, no system mode, no theme gallery, no layout picker.

Implements the "BloxOS Monoform Dark/Gray Reset" brief.

## What changed

- **One shell.** A 238px near-black rail and a single top bar own navigation, the page title, refresh, ⌘K, alerts, Add API, Add Machine and the account menu. All seven authenticated routes previously drew their own sticky `backdrop-blur-xl` header; all seven are gone.
- **One Overview.** `page.tsx` went 1120 → 727 lines, with seven presentational components extracted to `components/overview/`. Structure: intro → fleet posture grid → context grid → one authoritative **Machine fleet** table. `FleetPulse`'s bento of equal KPI cards is retired.
- **One appearance control.** A two-state segmented control in Preferences. `ThemeContext` exposes only `useTheme() → { appearance, setAppearance }`.
- **Deleted:** 16 client files and 2 hub files, including the four fleet layouts, `DesignContext`, `DesignSettings`, `ThemeSettings`, `ThemePreview`, the three `design-*.css` files and `/api/me/design`. `globals.css` went 729 → 153 lines (205 of those were named-theme blocks); `monoform.css` is the new 856-line token and component layer.

Preserved unchanged: refresh, command palette, alerts, add machine, API machine handling, search, filtering, saved filters, sorting, selection, bulk commands, edit/delete, machine routes, RBAC scopes, SSE behaviour, mobile support.

## Three latent bugs fixed in passing

None were introduced by this work.

1. **`--color-accent` was 14%-alpha everywhere.** It was defined twice inside one `@theme` block and resolved last-wins to `--accent-subtle`, so every `bg-accent` / `text-accent` in the app rendered as a faint wash — visibly wrong in `MachineCard.tsx:247` and the shell's primary button.
2. **A latent CSS cycle in the font tokens.** `@theme inline { --font-sans: var(--font-sans) }` emitted a self-referential `:root` declaration on the same element next/font targets, at equal specificity. It resolved only because `entryCSSFiles` happened to list globals first; reorder it and every Geist face plus the chart labels would silently drop to the system stack.
[...]

PR #227, opened 2026-09-10
Dark/light themes, minimal collapse, and software power measurement

bokiko:
Owner feedback after running Monoform on a real six-machine fleet, plus software power measurement end to end.

## Dashboard refinements

**Contrast modes are now `dark` (default) and `light`.** The name `gray` is gone from the product — a guard test asserts it can't return. Light is a real palette, not an inversion: `void` stays the recessed ground and `panel` stays raised, so elevation reads identically in both.

| token | light value | contrast vs canvas |
|---|---|---|
| `--mf-blue` | `#3149d9` | **6.18** |
| `--mf-green` | `#0d7346` | **5.36** |
| `--mf-amber` | `#8a5300` | **5.75** |
| `--mf-red` | `#c62828` | **5.11** |

Stored `"gray"`, `"system"` or a retired theme name resolves to `dark` on both client and hub. **No migration** — normalisation happens on read; `hub/migrations.go` untouched.

**Collapsing a section removes the section.** 288–348px → 30px: no box, no border, no padding, nothing reserving the old space. Chrome animates with the fold rather than snapping away on frame one. A folded pane keeps its column instead of letting its partner reflow — otherwise clicking one pane silently re-lays-out the one you didn't touch.

Also: all four Overview panes equal width with a shared height floor; the five other pages get the shortened intro; machine notes readable and editable from the Overview; and the expected-high-load control moved from a bare icon in the table's far-right actions column onto the Needs-attention row with a text label — a control that changes what the dashboard alerts on has to live where the alert appears.

### Measured — six sources, not one

```
system   rapl-psys → battery → ipmi-dcmi → hwmon:<chip>
cpu      rapl-package
dram     rapl-dram
```

`psys` is preferred over summing packages because it covers more of the platform — but only when the counter **actually advances during a 600ms probe**. Some boards expose the zone and never increment it, which would otherwise report a confident zero. `system` and `cpu` are separate fields, never summed.

### Estimated — only where nothing can be measured

Most ARM SoCs, VMs and Windows hosts expose no counter at all. RK3588 exposes regulator voltages but no current, so watts genuinely cannot be derived. Those machines can now report a **labelled** estimate:

```
P = P_idle + (P_max − P_idle) × u
```

`u` is CPU busy from `/proc/stat`, clamped not extrapolated, excluding `idle`, `iowait` **and** `steal` — an `iowait` CPU is halted at idle power, and `steal` is time this guest didn't consume. Linear on purpose: the true curve is concave, but that buys a few percent while the endpoints carry ±40%.

The table holds **whole-platform envelopes, not CPU TDPs**, keyed on device-tree `compatible` first. A TDP is a thermal limit for one component; the same chip in two boxes can idle five watts apart. **Unknown x86 reports nothing** — its PSU, drives and any discrete GPU can draw more than everything else combined. Unknown ARM falls back to a bounded per-core heuristic. `BLOXOS_POWER_ESTIMATE_WATTS="<idle>:<max>"` overrides both; a malformed value **disables** estimation rather than guessing.

**The estimator engages only when `system`, `cpu`, `dram` and GPU are all unavailable**, `nvidia-smi` included. Not for tidiness: a wrong envelope can put estimated `system` *below* measured `cpu` — visibly impossible, and it discredits every other number the agent ships. With a measured 300 W card the contradiction is far larger.

**No `estimated` boolean on the wire.** The hub stores buckets by decode-then-re-encode under its own schema, so a flag an older hub doesn't know is silently dropped — and a dropped `estimated` reads as `false`, meaning *measured*. That fails open on the one property that must never be wrong. `sources` cannot vanish that way; a proto test proves the label survives a round trip through a 3a-shaped hub.

### Fleet panel

`GET /api/fleet/power/history` returns measured and estimated as **separate series** with explicit coverage, so a total drawn from 3 of 6 machines cannot be misread as a fleet total. The Overview pane charts it with a configurable cost rate, replacing the empty "Capacity over time" placeholder.

### Accuracy, stated plainly

| Basis | Expected error |
|---|---|
| Table hit, CPU-dominated | ±30–40% |
| Per-core ARM heuristic | ~2× |
| Operator override from a wall meter | ±5–10% |

Trend-grade, not billing-grade. Documented worst cases: uncounted accelerators, memory-bound work, and thermal throttling holding utilisation at 100% while real power falls.

## Protocol compatibility

Strictly additive — `System`, `DRAM`, `Sources` all `omitempty`.

- **Old agent → new hub.** Fields absent decode as *unavailable*, never zero, and are never given an invented label. Agents that cannot be updated keep reporting indefinitely.
- **New agent → old hub.** The old hub drops what it doesn't know; the replay-conflict check still works because it normalises *both* sides through its own schema.
- **Newer agent → this hub.** Unknown domain labels are bounds-checked and stored rather than rejected, so upgrading the hub is never a precondition for upgrading an agent.

## Verification

```
dashboard  pnpm test   152 pass, 0 fail   (was 107 at v1.4.0)
dashboard  pnpm lint   0 errors, 3 warnings (pre-existing)
hub        go test     ok (112s, -count=1)
proto      go test     ok
agent      GOOS=linux + GOOS=windows build clean; suite run in a Linux VM
gofmt      clean beyond the 4 files already unformatted on main
```

Rendered in a browser against a live fleet in both themes. `hub/migrations.go` untouched.

## Known issue, not fixed here

BloxOS reports **core 0's CPU model × total core count**, so an RK3588 shows as "8× Cortex-A55" when it is really 4× Cortex-A76 + 4× Cortex-A55. This is wrong on every big.LITTLE board. Surfaced while building the estimator — which is why its table keys on device-tree `compatible` rather than the CPU string. Tracked separately.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 5893a3fcd1497bf1e95d78d30b87513c3d79ee2a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bokiko (commented):
## Review — needs-work

Reviewed tip `b0e0fc284dc4eec371d5b8537fb4b143a1587bd0`. Did not merge.

### Blockers

1. **Settings chrome still broken (same as #226 needs-work)** — `dashboard/src/app/settings/page.tsx` still has `data-design-page` + sticky `backdrop-blur-xl` header + Fleet back link inside `AppShell`. This PR only retints `TabsList` to `bg-surface-sunken`. Please remove the sticky blur chrome and `data-design-page`, and let AppShell own navigation/title (optionally `usePageTitle("Settings")`).

2. **`useMachineNotes.ts` is Git-binary (`+0/−0`)** — `KEY_SEP = "\0"` (literal NUL) makes the file unreviewable in the PR diff. The hook itself is fine (~fan-out with concurrency 4, documented cost). Replace with a printable separator (`"\x1f"` / `|` / similar).

### P2 (non-blocking but real)

- **Battery discovery** (`agent/power_sources_linux.go`): comment says judge by device/files existing, not discharge state; code requires a non-zero watt reading at detect time. Full/idle at agent start can miss the battery backend until restart.
- Extend `monoform-guard` to ban `data-design-page` so Settings cannot keep regressing.

### Otherwise looks good

- **Theme:** dark|light bootstrap / ThemeContext / hub `user_prefs` align; gray/system/junk → dark; light removes `.dark`; guards ban gray; `migrations.go` untouched.
- **Collapse:** Disclosure + `.mf-section` / pane grid — folded ≈ hairline, chrome animates with content, partner column preserved.
- **Power:** ordered backends + psys 600ms probe; domains not summed; mixed window drops domain; proto omitempty + compat; hub old-agent / unknown-label behavior sound.
- Expected-high-load on Needs-attention; workspace prefs local-only.

Happy to re-review once Settings + the notes NUL are fixed. COMMENT only — do not merge.

PR #232, opened 2026-09-12
fix(power): stop presenting sampled power as measured energy and current draw

bokiko:
PR1 of the power-correctness work. **Server + dashboard only.** No agent, release, updater or delivery changes.

## The three claims this removes

**Energy was never a floor.** `hub/fleet_power.go`'s own header asserted "every energy figure is a FLOOR", and the pane rendered `At least $X · Y kWh` on that basis. A window's mean is the mean of the reads that *succeeded* in it, and the hub then weights that mean by the window's whole span — one successful 300 W read in a 30 s window contributes 9000 W·s as though all thirty seconds were observed. The caption beside it divided summed window spans by the period and called that coverage, but spans are not deduplicated and a window binned by its end can carry time from the previous bucket.

Energy and cost are **withheld**, not re-captioned. Nothing renders `0 kWh`: unavailable accounting is not zero energy.

**"Current" came from the chart.** The headline walked back through the charted buckets and printed the last non-null one — so it changed with the selected period, reached arbitrarily far back, and was aged against bucket edges that can sit in the future. A fleet dark since 02:00 still read "142 W measured" at noon. New `GET /api/fleet/power/current` answers that from each machine's newest window over a fixed lookback, gating every contributor individually.

**Unrecognised backends became measurements.** Classification was "estimated if `IsEstimatedSource`, else measured", so any label this hub didn't know was promoted to measured. It now fails closed: `unknown` is its own kind, excluded from both totals and counted so the exclusion is visible. Legacy unlabelled readings stay measured — those agents predate labelling and measured RAPL.

## Two bugs found while testing

- A window covering `[30s,60s)` was binned into the **next** bucket.
- The SQL predicate excluded **records ending exactly at the requested end**. Such a record lies inside the requested window, so where one existed it was dropped from the newest bucket. (It does not affect every response — a request end can sit in the future, in which case no record ends exactly on it.) Fixing the aggregator alone would not have helped: rows the query never returns cannot be re-binned.

## Completeness

`complete` now reports false and says why — spans aren't enforced non-overlapping at ingest, binning by end can credit a neighbouring bucket, and a window's mean comes only from its successful reads. Presence is exposed separately as `all_machines_contributed`.

## Domain choice

Was automatic: system power took the whole chart whenever any machine had it. Once a *modelled* system reading could appear, one estimating board would have evicted the measured CPU and GPU history of the entire fleet. Choice is now explicit, persisted, and latched once.

## Verification

- hub `go vet` + full suite exit 0; proto vet + tests exit 0
- dashboard 191 tests pass; lint exit 0 at the documented 3-warning baseline; build compiles
- rendered at 1440 and 1280 in both contrast modes against a local stub hub

### Independent verification (reviewer)

Re-ran hub `TestFleetPower*`, dashboard targeted tests and lint (pass, baseline 3 warnings), and rendered with isolated headless Playwright at **explicit 1440 / 1280 / 390 viewports in both dark and light** — all six renders confirmed at the exact viewport, with no clipped power-panel buttons, `scrollWidth == clientWidth`, the saved CPU selection preserved across reload, and no runtime or hydration errors. PNGs inspected.

Artifacts: `/private/tmp/bloxos-browser-review/{power,page}-{1440,1280,390}-{dark,light}.png`, with `results.json` and `review.cjs` alongside.

This resolves the 390 px viewport gap noted during authoring, where Chrome's interactive window floor (~519 px) prevented a true narrow render.

## Known limitations

- Energy/cost stays withheld until observations carry their own measured duration — additive protocol work, deliberately not here.

---

> [!NOTE]
> **Medium Risk**
> Changes fleet power semantics, aggregation, and a new authenticated API that the overview depends on; behavior is more conservative but operators will see different numbers and no energy/cost line until a follow-up.
> 
> **Overview**
> **Fleet power is reframed so the UI and hub stop overstating what sampled telemetry can prove.** Energy and cost are **withheld** on the overview (no “at least” kWh/$); the hub withdraws the “energy is a floor” claim and exposes sample/window metadata instead.
> 
> **“Current” draw is no longer derived from chart history.** The dashboard calls new **`GET /api/fleet/power/current`** for headline readouts, with shared **`power-freshness`** rules (age from agent window ends, aggregate aged by oldest contributor). Stale, skewed, unreadable, and unclassified machines are called out rather than folded into totals.
> 
> **Domain view is explicit:** a segmented control (persisted in `localStorage`, latched after mount) replaces automatic “system wins” mode; series combine history and the current snapshot so live domains are not hidden by capped history.
> 
> **Hub aggregation fixes and stricter provenance:** readings are binned on `(end-1)` and history SQL uses `(start, end]` so edge windows are not dropped; backend labels **fail closed** to an `unknown` kind (legacy unlabelled stays measured); `complete` is always false with **`all_machines_contributed`** for presence; clock-skew tolerance is tightened for freshness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a79b632a004da2f4066f11c9ac1c6a90ac2426e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dashboard/src/components/overview/FleetPowerPane.tsx
+++ b/dashboard/src/components/overview/FleetPowerPane.tsx
@@ -295,9 +295,10 @@ export function FleetPowerPane({ period, onPeriodChange }: FleetPowerPaneProps) 
 
   // Series come from BOTH sources: a capped or empty history must not hide a
   // domain that is reporting right now.
+  const current = state.current;
   const series = useMemo(
-    () => (combinedSeries(data, state.current, selected) as Series[]) ?? [],
-    [data, state.current, selected],
+    () => (combinedSeries(data, current, selected) as Series[]) ?? [],
+    [data, current, selected],
   );
   const rows = useMemo(() => fleetPowerChartRows(data, series), [data, series]);
   const warnings = useMemo(() => (fleetPowerWarnings(data) as string[]) ?? [], [data]);
--- a/dashboard/src/components/overview/FleetPowerPane.tsx
+++ b/dashboard/src/components/overview/FleetPowerPane.tsx
@@ -40,28 +40,30 @@
  * green/amber/red appear here only for a real warning, with words.
  * ========================================================================== */
 
-import { useEffect, useMemo, useState } from "react";
-import Link from "next/link";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { CartesianGrid, Line, LineChart, ResponsiveContainer, Tooltip, XAxis, YAxis } from "recharts";
 import { ChartTooltip } from "@/components/charts/ChartTooltip";
 import { Zap } from "lucide-react";
 
 import { DEMO_MODE, HUB_URL, getStoredToken } from "@/lib/session";
+import { freshnessNote } from "@/lib/power-freshness.mjs";
 import {
   PERIOD_LABELS,
   POWER_PERIODS,
   coverageSentence,
   domainLabel,
-  domainOf,
   fleetPowerChartRows,
-  fleetPowerCost,
-  fleetPowerMode,
-  fleetPowerSeries,
+  fleetPowerEnergyAvailability,
   fleetPowerWarnings,
-  formatKWh,
-  formatMoney,
   formatWatts,
-  latestReading,
+  currentReading,
+  currentDomainOf,
+  domainOf,
+  domainSeries,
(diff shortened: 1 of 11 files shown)


bokiko:

PR #233, opened 2026-09-12
feat(delivery): ship agent payloads with the hub release [DRAFT]

bokiko:
> **DRAFT — incomplete, no merge requested.** Opened early, on Codex's instruction, so Docker/Compose and native CI start running against the packaging changes while the remaining work lands. Do not review as a finished change.

Based on `fix/fleet-power-honest-claims` (PR #232) while that is unmerged, so the diff shows only delivery changes. Will be retargeted and rebased onto `main` once PR1 lands.

## The problem

`bloxos-update` replaces the hub and dashboard and deliberately does not touch agent files. A v1.7.1 hub therefore served agents built weeks earlier, and every machine reported "matches offered build" — correctly. The offer was stale. That single fact accounted for missing power on both RK3588 boards, absent source labels fleet-wide, and wrong big.LITTLE CPU inventory.

**Placement alone would not have fixed it.** The resolver already had a `hub-executable-directory` candidate sitting *last*, behind the system defaults, and the project's own systemd unit sets `BLOXOS_AGENT_BINARY`, which for amd64 is the authoritative fail-closed *first* candidate. Shipping agents in the release without the precedence contract would have changed nothing.

## What is here so far

**Resolution and validation.** The hub discovers a bundle at `<hub executable dir>/agents/`, and a present-but-invalid bundle is fatal rather than a fallback to frozen defaults — falling through is precisely how a hub ends up quietly serving months-old agents while reporting success. Delivery-config errors are retained on the resolver and fail every resolution; the startup gate is the first statement of `main()`, before the database opens. Catalog identity is re-checked on every resolution, and each payload's embedded release marker is cross-checked against the catalog.

**Architecture is read from the image, not the label.** A sha256 proves a payload is the file the catalog names; it says nothing about what those bytes *are*. Every declared platform is now verified at load — `verifyELFArch` for Linux, a new `verifyPEArch` for Windows. The PE header offset is read out of the file under test, so it is bounded before being used as a seek target.

**The manifest is trusted like a payload.** It was `Stat`'d and then read. `Stat` follows symlinks; a FIFO stats as zero bytes, passes the size limit, and then blocks forever — confirmed, the previous loader hangs on one indefinitely. It now goes through the same trust check as the payloads it describes, with the bound enforced by the read.

**Packaging.** Both server archives carry all three agent payloads plus the catalog under `hub/agents/`. The bytes are copied from the same directory published as standalone assets and asserted equal, never rebuilt. **Nothing in the updater changes** — `safe_extract_tar` has no allowlist and `stage()` copytrees the whole extracted tree — and there is no manifest schema bump, which would have forced workers to be upgraded before they could carry the thing that upgrades them.

**One layout for both deploy modes.** `agentBundleRequired` was set by nothing, so the "a packaged build must have a bundle" path had never been reachable. It could not simply be added, since the archive's hub *is* the Docker binary and containerised hubs had no bundle. Both modes now use `agents/` beside the hub executable, with the legacy `/usr/local/lib/bloxos` paths as **hard links** to the managed payloads — regular files, so the resolver accepts them, and the same inode, so they cannot drift from the managed bytes.

**In-image catalog omits what it cannot know.** The image digest does not exist until the image exists, and a local `docker build` has no tag or commit SHA. Those fields are left out rather than filled with a placeholder: a fake all-zero SHA is a field that looks like evidence and is not. Artifact checks stay mandatory in both cases; release catalogs still require full provenance strictly.

## Testing

`scripts/test_server_bundle_agents.py` packs, transports and loads with the real components: the real packer, the unchanged updater (both `safe_extract_tar` and `NativeAdapter.stage`, the path a host actually takes), and the hub's own Go loader over the resulting tree — asserting it PASSED rather than skipped.

That cross-component test immediately earned itself: it caught the packer's ELF check accepting an `EI_VERSION` the hub's loader rejects. Real Go builds set `EV_CURRENT`, so the packer was aligned to the stricter reading rather than the hub loosened — validators that disagree let a release pass packaging and fail at hub startup.

Every rejection test is paired with a control that must still load, and the five loader-hardening tests were confirmed to fail against the previous loader.

## Not done yet

- Release identity gate (`agentRelease` bump, published `(platform, release) → SHA` immutability before promotion)
- Content-addressed signature lookup
- Durable automatic staged rollout — planned as a dependent follow-up PR so this artifact-delivery change stays independently reviewable
- Docs and the Versions UI

**Neither PR is a release on its own.** Merging this alone is not completion of the agent self-update work.

## Verification gap, stated plainly

Docker is not available on the machine this was written on, so the image build is **unverified locally** and this CI run is its first real test. That is why the draft is open now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,19 @@ jobs:
       - name: Test agent payload packaging and old-worker transport
         run: python3 ../scripts/test_server_bundle_agents.py
 
+      # Runs the real hub PROCESS. A loader unit test cannot catch the gate
+      # being moved after the database opens, and the fixtures need root-owned
+      # paths because the trusted-binary rule demands root ownership of every
+      # ancestor. REQUIRE_ROOT makes a missing prerequisite fail rather than
+      # skip: a skip reads exactly like a pass in the checks summary, and this
+      # file exists to catch a check that has quietly stopped running.
+      #
+      # This form proves the gate is wired and reachable. It does NOT verify
+      # the release build sets the marker — the flag is supplied here. The
+      # image-built hub is covered by the Compose smoke job.
+      - name: Hub delivery boot gate (source build)
+        run: sudo -E env "PATH=$PATH" BLOXOS_BOOT_SMOKE_REQUIRE_ROOT=1 python3 ../scripts/test_hub_bundle_boot.py
+
   # A real gate as of issue #60: the hub's DB handle and agent registry now live
   # on a Server instance, so tests build isolated servers instead of reassigning
   # package globals under leaked agent-WS goroutines. Verified stable across four
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,14 @@ jobs:
           python3 ../scripts/test_agent_bundle.py
           bash -n ../scripts/export-agent-bundle.sh
 
+      # Runs in THIS job because it needs both toolchains: the packer is
+      # Python, and it drives the hub's real Go loader over the tree an
+      # unchanged updater produced. Deliberately before publication — a
+      # tag-time test finds a broken release after the tag exists, and the
+      # fleet is who finds out.
+      - name: Test agent payload packaging and old-worker transport
+        run: python3 ../scripts/test_server_bundle_agents.py
+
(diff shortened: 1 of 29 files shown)


bokiko:

PR #234, opened 2026-09-12
feat(rollout): durable automatic staged agent rollout

bokiko:
Draft, stacked on #233 (PR2a, frozen at `d8a3eca`). Not approved, not merged, not tagged, not deployed.

## Before

The hub announced a new agent build to **every connected agent at once**, driven by whatever happened to trigger it — a WebSocket connect, a version report, a restart. Nothing was written down. If the hub restarted mid-rollout it had no idea what it had already offered; if a build was bad, the fleet had already taken it. A fleet-wide circuit breaker paused *everything* after a couple of failed reconnects — one bad machine stopped every platform, and recovery was a manual reset of in-memory state that no restart preserved.

## After

One scheduler goroutine owns sending. Triggers only call `wakeRollout()`; none of them reserves or writes. Admission happens on **capacity alone**, so a machine held back because the batch was full is picked up by a later pass with no further trigger — requiring one looks identical to working right up until a fleet stops updating because nothing poked it.

**Stages.** Per platform (`linux/amd64`, `linux/arm64`, `windows/amd64`), independently: one canary, then batches of two. A stage advances only when nothing in it is still outstanding **and** at least one machine in it validated — without the second condition, a stage whose every candidate was withheld would advance having learned nothing about the build.

**Durable reservations.** A slot is taken in SQLite *before* anything reaches a socket. The capacity check, the stage read and the insert are one transaction under `_txlock=immediate`, which is what bounds concurrency between *different* machines — a unique key on (machine, candidate) only deduplicates repeat calls for the same machine, and two machines racing for the final slot would both have inserted.

**Generations.** A changed candidate bumps the platform's generation; every slot, reservation and piece of evidence is scoped to one. A report that arrives just after the served binary changed cannot mark the new generation's slot observed on the strength of the machine running the old bytes.

## Crash semantics: at-least-once, and honest about it

A process that dies between reserving a slot and recording the offer leaves **the same row** whether the write landed or not. The hub does not guess. Recovery is a bounded resend under the same slot identity — the deadline and grace are *preserved*, because refreshing them would let a crash loop buy unlimited time — and the ambiguity is resolved by **what the machine reports when it comes back**. If it reports the candidate, the slot becomes observed; if it reports the old build, it is resent.

Delivery is at-least-once and is not claimed otherwise. Accordingly, an expired reservation reports *"update was not confirmed within the attempt window"* rather than accusing the machine of refusing an offer it may be running.

## Health

**A matching SHA is not health.** Version reports establish which bytes; metrics establish that the thing is alive. A machine is validated by 60 seconds of continuous telemetry with inter-frame gaps bounded at 45s, and the dwell endpoint must be a real frame — comparing the clock would complete a dwell during silence.

**Evidence belongs to a connection, not a machine.** `agentRunningVersions` survives socket replacement, so a fresh connection would otherwise inherit the previous one's matching SHA. Records are keyed by `*ConnectedAgent`, held in memory only, and dropped as each handler exits. A reconnect proves itself again.

This is a liveness check. It does not prove any feature still works, and the docs say so in those words.

## Operator controls

The operator pause and a platform halt are **different things**, and the page now says which is which:

| | Operator pause | Platform halt |
|---|---|---|
| Set by | a person | the hub, on a failed attempt |
| Scope | fleet | one platform |
| Survives restart | yes | yes |

`rollout_paused` means only that someone pressed pause. **Retry halted rollouts** appears whenever any platform is halted — previously the only recovery control appeared when the pause was *on*, so a halted platform with the pause off offered a Pause button and nothing else. It is the existing `POST /api/versions/resume`, which clears the pause and gives every halted platform a fresh attempt in one transaction; the wording states that fleet-wide scope because there is no per-platform retry.

The pause is honoured **at the send boundary**, not only where the announcement was decided, and so is registry ownership: a goroutine can queue behind unrelated socket writes, and in that window the machine may be deleted or taken over by a new connection.

**Deletion.** `agent_rollout_slot` is deleted inside the machine's own delete transaction, so a rollback keeps its rollout state. An ingestion barrier held across the commit stops a queued pass inserting a slot for a machine that is gone.

## What the UI claims

Counts are rendered through a pure, separately tested module. Each platform row names the candidate SHA its counts describe — the hub only picks up a new candidate when a reservation runs, so with the pause on it can serve one build while the rollout still describes the previous one.

- *"N observed on this build"* — the underlying flag is a latch; the machine may since have failed or rolled back.
- *"N pending attempts"* — reservation precedes the socket write, and a resume opens an attempt for machines already on the build, which are never announced to.
- *"no outstanding tracked attempts"* — it counts slots, which exist only for machines this generation reserved one for, so it cannot speak for the connected fleet.
- An unrecognised or missing status renders **Unavailable**, never a healthy zero.

## Validation

Full hub suite exit 0 (118.6s), hub vet 0. Dashboard `tsc` 0, lint 0 at the 3-warning baseline, 209 node tests. The reviewer independently ran the three send tests under `-race` (7.7s) and verified the page in a browser: clicking Retry with the pause off sends exactly `POST /api/versions/resume`; the unavailable sentinel renders Unavailable with its reason and no retry offered; 1440/1280/390 in both themes, no errors or overflow.

Crash-window coverage is **fault injection**: the announce is aborted after the durable reserve and before any write, and after a successful write and before the offer is marked; the controller is then rebuilt over the same database and the machine reconnects on a new socket.

## Limitations

- Delivery is at-least-once, not exactly-once.
- Validation is liveness only.
- Health evidence is in-memory, so a restart resets dwells that were **in flight**. Slots already validated stay validated — they are durable.
- Full publish-retry idempotency is **not** addressed here and is not claimed — that is PR2c, which is required before any release.

### DIFF
--- a/dashboard/src/app/versions/page.tsx
+++ b/dashboard/src/app/versions/page.tsx
@@ -20,7 +20,7 @@ import {
   Package,
   ShieldCheck,
 } from "lucide-react";
-import { AgentBinaryInfo, useVersions } from "@/contexts/VersionsContext";
+import { AgentBinaryInfo, AgentRolloutStatus, useVersions } from "@/contexts/VersionsContext";
 
 /**
  * The resolution POLICY in force — not a claim about any particular platform.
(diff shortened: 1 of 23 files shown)


bokiko:

PR #235, opened 2026-09-12
fix(release): deterministic assets and a publish retry that reuses what it verified

bokiko:
Draft, stacked on #234 (PR2b) which is stacked on #233 (PR2a, frozen). Not approved, not merged, not tagged, not deployed.

**Required before any release.** Both problems here are in the release pipeline itself, and both make a failed publish unrecoverable.

## 1. Release assets were not a function of their content

Two builds of the same revision produced different bytes, so "retry the publish" could never be proven to re-produce what it was retrying — which is the foundation the whole no-clobber policy stands on. Every cause was metadata:

| Source | Why it leaked | Fix |
|---|---|---|
| gzip header | `tarfile.open(path, "w:gz")` hands the **output path** to gzip, which writes it *and* the compression time into the header | build the stream explicitly, `mtime=0`, empty filename |
| tar members | mtime, uid/gid, uname/gname — an archive recording `runner:docker` describes the machine that built it | normalise at the serialisation boundary; mode reduced to "is it executable" |
| zipapp entries | each source file's mtime and mode, from a copy into a temp dir moments earlier | fixed entry metadata, same layout, stored uncompressed |

The mtime is an **integer** deliberately: a float makes `tarfile` emit a pax header to carry the fraction, so the member's encoded length would depend on the filesystem's timestamp resolution.

Member *order* needed no intervention — `tarfile.add` and `zipapp` both already walk sorted. Nothing about what the archives contain changed. Normalising on the way out is also why the staging tree's own timestamps and path stop mattering.

**Test:** the real entrypoint runs twice with the clock, the staging tree's timestamps, the umask and every path made to differ; all output bytes must match — both archives, the updater zipapp and the manifest. It also asserts the published standalone agent payloads were not touched, and that the unchanged updater still carries the result, so determinism cannot be bought by breaking transport.

## 2. A publish retry could not finish

A retry rebuilds both images, and nothing guarantees a rebuild lands on the same index digest. When it does not, the catalog derived from the new hub image differs from the one the first attempt already persisted — and a published catalog is never overwritten, so the retry **fails at that comparison and stops**.

The ordering from PR2a holds and does its job here: the catalog is persisted *before* promotion, so nothing mismatched ever reaches the Docker tags. The damage is narrower and still fatal — that release can no longer be completed by any number of retries. The no-clobber policy was rejecting its own recovery path.

`release-build.json` is a small immutable release asset — schema, version, revision, both canonical digest-pinned refs. Looked up **before** the builds:

- positively absent → build, as now
- present and matching → **skip both builds, reuse the refs**
- unreadable, wrong schema, different source, or a **failed lookup** → fail closed

Absence is the only outcome that means "build"; treating a rate limit or an expired token as absence would reintroduce exactly the divergence this prevents.

The record is persisted **after** identity validation and **before** the catalog or any tag moves. A crash after it leaves a verified-but-unpromoted pair, which is what the retry needs; a crash before it leaves nothing, so the retry simply builds. Reused images are revalidated exactly as freshly built ones — same extraction, same per-platform payload checks, same identity gate.

The updater's manifest schema is untouched. This is a CI-only record; no worker reads it.

## Also fixed

- **Both images are validated before anything is recorded or promoted**, on both architectures. The export path checked the dashboard image's revision only *after* promotion, and a record matching this run's sha is evidence about the record, not about the images it names. Uses the function the export already uses, lifted out and shared.
- **`docker pull` writes progress to stdout**, and the helper is invoked as `ref=$(... --revision ...)`, so progress was spliced into the reference handed back to docker. Now on stderr.
- **Two shell branches read `gh release view` failing as "no release exists"**, so an auth or network failure would try to create an already-published release and skip the assertion that it is still a draft. Both now go through one listing that either loads or raises.

## Validation

A reuse path that resolves the recorded pair and then promotes `steps.hub.outputs.digest` anyway is inert on the happy path and wrong on the one run that matters, because a skipped build's output is the empty string. So the wiring is tested two ways:

- **Structural** — no step outside the resolver reads a build digest; the job's `outputs:` block maps `hub_ref` → `steps.refs.outputs.hub` exactly; both builds carry the reuse guard; the ordering holds; neither job branches on a failed release lookup.
- **Behavioural** — the actual `run:` blocks *and* the actual `env:` mapping are lifted from the workflow, resolved against fixtures, and executed under bash with a fake docker that logs its argv. Fresh run, reuse run (build outputs empty, recorded refs deliberately different from the built ones), and a crash injected after the hub tag is live.

Every rejection test was run against the broken version and fails for its intended reason — including the two that motivated the harness: swapping `HUB_REF`/`DASHBOARD_REF` in the step env promotes hub tags to the dashboard digest, and a reuse branch that falls through to the build outputs yields `ghcr.io/bokiko/bloxos-hub@` with no digest.

## Suites

release-history 26, export 13, workflow 12, agent-bundle 32, release-identity 22 — all pass. `scripts/test_release_workflow.py` is registered in `ci.yml`.

## Limitations

The retry path is tested against the workflow's own shell with a fake docker. It has **not** been exercised against a real tagged release, and will not be until a release is authorised.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,10 +89,11 @@ jobs:
       # This form proves the gate is wired and reachable. It does NOT verify
       # the release build sets the marker — the flag is supplied here. The
       # image-built hub is covered by the Compose smoke job.
-      - name: Test release identity gate and history
+      - name: Test release identity gate, history and workflow wiring
         run: |
           python3 ../scripts/test_release_identity.py
           python3 ../scripts/test_release_history.py
+          python3 ../scripts/test_release_workflow.py
 
       - name: Hub delivery boot gate (source build)
         run: sudo -E env "PATH=$PATH" BLOXOS_BOOT_SMOKE_REQUIRE_ROOT=1 python3 ../scripts/test_hub_bundle_boot.py
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -166,8 +166,11 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     outputs:
-      hub_digest: ${{ steps.hub.outputs.digest }}
-      dashboard_digest: ${{ steps.dashboard.outputs.digest }}
+      # The RESOLVED pair, which on a retry comes from the record a previous
+      # attempt published rather than from a rebuild. Everything downstream
+      # reads these, so a skipped build cannot leave a dangling reference.
+      hub_ref: ${{ steps.refs.outputs.hub }}
+      dashboard_ref: ${{ steps.refs.outputs.dashboard }}
     permissions:
       # contents: write is needed to persist the verified agent catalog to the
       # draft release BEFORE the Docker tags are promoted.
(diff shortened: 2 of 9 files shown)


bokiko:

PR #236, opened 2026-09-12
fix(power): measure or say nothing

bokiko:
Draft, stacked on #235 (PR2c) → #234 (PR2b) → #233 (PR2a). Not approved, not merged, not tagged, not deployed.

Governing rule, from `AGENTS.md`: **never estimate — a machine with no counter reports nothing.**

## Estimate generation removed

`power_estimate.go` and its Linux half are deleted — profiles, the per-core SoC curve, the envelope override, the utilisation sampler, and the `/proc/stat` reader that fed it. `BLOXOS_POWER_ESTIMATE_WATTS` and `BLOXOS_POWER_ESTIMATE` are gone.

**Decoding is retained.** Rows written by older agents sit in journals and replay after an upgrade; those agents also still report. `SourceEstimateUtil` and `IsEstimatedSource` remain, a modelled row replays labelled, and the hub presents it as modelled — separately, never summed. A journal test writes such a row and reopens it under a build containing no estimator.

## RAPL

**Per-counter progress.** Every participating counter must advance on its own account. One busy socket used to certify a frozen sibling: the sum stayed plausible and shipped as the whole domain. With `minWatts` 0, an entirely frozen package or dram group produced **0.0 W and passed** — a running CPU reported as drawing nothing.

**Backward steps produce a gap, not a correction.** A rollover and a counter reset (suspend/resume, driver reload) are indistinguishable from two reads, and assuming a wrap invents up to a whole `max_energy_range_uj` of energy at a rate the plausibility ceiling accepts. Any backward step is unavailable and re-primes; the next interval recovers on its own. The declared range is validated before subtraction.

**Discovery withholds rather than skips.** A zone whose *name* could not be read was skipped, so on a two-socket board with `package-1` unreadable the CPU domain reported one socket as the machine. Such a zone is not absent — only its domain is unknown. A top-level one could be package or psys and withholds both; a sub-zone could only be dram. Named core/uncore sub-zones are not counted as missing dram contributors.

**A VM guest withholds every RAPL domain** — `cpu` and `dram` as well as `system`. A package counter measures physical silicon the guest shares with tenants it cannot see. A virtualization host keeps all three. Detection is gopsutil's, injectable; no `systemd-detect-virt`, no CPUID or DMI heuristic.

## Two system backends removed

Neither established scope.

- **Generic hwmon**, every chip including `power_meter`. A shunt reports whatever rail it is wired across; the kernel's documentation gives `power*_is_battery` and the `measures/` symlinks precisely because the chip name does not say. Device-scoped DRM GPU hwmon is untouched.
- **Battery.** A pack's discharge is the machine's draw only while it runs on that pack alone. Discovery also dropped a pack with no usable reading before building the sampler, so a two-pack machine became a one-pack undercount; and a charging sibling contributed nothing while the reading still counted as a system total.

Remaining: **RAPL psys**, and an in-band DCMI reading the BMC reports as active.

## IPMI DCMI

`ipmitool` prints the instantaneous value and **exits zero** even when the BMC reports the reading state as `deactivated`, so a number plus a successful exit was becoming measured whole-system watts. The state line is now required. The value line is matched whole with units — loosely anchored, `212garbage Watts` yielded 212 and `1e6 Watts` yielded 1 — and duplicate conflicting lines are refused.

**An active zero survives.** `powerSystemMinWatts` is the energy counter's tie-breaker; a direct reading has no tie to break.

## Hub read time

Removing a backend does not remove its rows. Excluded from **system** totals, current and historical, and folded into the existing unknown/excluded count — no new category, no wire field, no schema change, stored rows unmodified:

- battery and `hwmon:*` in the system domain;
- an all-zero RAPL window. Not a claim the row is wrong: samples and expected cannot separate a frozen counter from genuinely zero energy, and that ambiguity is the reason to withhold. Not a `> 0` filter — a GPU at 0 W, an active BMC zero, and a window whose peak moved all survive.

Two further classifier assumptions were corrected against the git history:

- **The unlabelled exemption is CPU-only.** `System`, `DRAM` and `Sources` were introduced in one commit, so no build ever emitted an unlabelled System or DRAM reading.
- **A source must match the domain it measures.** The flat known-measured set accepted `rapl-package` as a system reading and `ipmi-dcmi` as a CPU reading. Each label is now bound to its domain; a mismatch is unknown. Component CPU/DRAM hwmon compatibility is retained.

Caption: **"N machines whose power source or scope is unverified — excluded from both series"**, with a tooltip distinguishing the two reasons.

## Validation

Full hub suite exit 0 (123.4s), hub vet 0. Full agent suite passes on Linux, agent vet 0 for `GOOS=linux`, gofmt clean. `proto/powerhistory` green. Dashboard `tsc` 0, lint 0 at the 3-warning baseline, 209 node tests.

The agent's `_linux.go` files cannot run on macOS, so the suite is cross-compiled for `linux/arm64` and executed in a Linux VM.

## Limitation

Validated with simulated sensor files, parser fixtures and Linux tests; no live-fleet hardware validation or deployment was performed. Availability checks do not certify sensor calibration.


bokiko:

PR #237, opened 2026-09-12
Fleet Power column, machine Overview power, and one current snapshot behind both

bokiko:
Stacked on PR3 #236 at `c522d27`.

Adds a Power column to the fleet table and the same readings to Machine Overview. System, CPU package, GPU total and DRAM remain separate. Values are labelled as 30-second sample means; modelled values carry both `~` and "Modelled". Missing, stale, skewed and unverified readings stay unavailable, while valid measured zero survives. Machine tags move under the hostname.

The existing current-power endpoint now includes per-machine results from the same decisions used for fleet totals. One shared provider supplies the table, Overview and fleet headline, with a hub-based clock, independent polling and session-bound state. A current-request failure clears current numbers; a history failure does not hide a valid current snapshot.

The machine Power history view now applies the shared provenance policy, separates measured and modelled series and summaries, and breaks lines across unavailable windows or backend changes. Isolated readings remain visible. A narrow versions-array guard also prevents older or incomplete versions responses from crashing card view.

Validation:
- Full hub suite passed (124.2s), hub vet passed; changed Go files are formatted.
- 261 dashboard tests passed; TypeScript, lint and production build passed. Lint retains three baseline warnings.
- Shared classification fixtures are read by Go and JavaScript tests.
- Browser checks at 1440, 1280 and 390 pixels in both themes covered measured/modelled/zero/unavailable states, request failures, browser clock skew and delayed responses across account changes.

No live-fleet validation or deployment was performed. Energy and cost remain unavailable. PR5 handles the fleet chart's visual finish. Three untouched hub test files retain pre-existing formatting issues.


bokiko:

PR #238, opened 2026-09-12
Fleet power chart finish, and an excluded domain the selector could not reach

bokiko:
Stacked on PR4 #237 at `db93da9`.

Finishes the fleet power chart's presentation and fixes one selector hole it exposed.

Chart: 2.25px strokes with round caps and joins, straight segments between sampled means, a quieter grid, and a taller plot area. Measured stays solid and modelled stays dashed; the tooltip now marks the modelled value with `~` as well as naming the series. Buckets with no neighbour are drawn as neutral marks instead of disappearing — a single-bucket window previously rendered an empty chart. The marks carry no freshness meaning, because a bucket's end time is not an observation time.

Controls: the domain and window selectors are one group that wraps at narrow widths, so every option stays visible rather than scrolling out of view.

Selector: a domain whose current contributors were all excluded is now offered. Previously only domains with a contributor appeared, so the hub's recorded exclusion reason was unreachable. Selecting such a domain shows the exclusion counts and an empty chart — no series and no numeric zero. The empty state no longer pre-empts that explanation when the history request has also failed.

The existing pane guards are unchanged and still pass: no area, gradient, shadow, blur or hardcoded hex; domains are never summed; energy and cost remain unavailable.

Validation:
- 262 dashboard tests passed; TypeScript, lint and production build passed. Lint retains three baseline warnings.
- Browser checks at 1440, 1280 and 390 pixels in both themes: every control inside its own group, computed stroke width, cap and join, no curve commands in any path, modelled dashed, isolated marks drawn, modelled tooltip marked, no page errors.
- The exclusion-only case is verified against the reproduction that exposed it, and controlled against the previous behaviour.

No hub or agent changes. No live-fleet validation or deployment.


bokiko:

PR #250, opened 2026-09-19
fix(power): read the full fleet history window

bokiko:
The 24-hour fleet chart stops reading after 20,000 raw windows, so an ordinary fleet can lose the oldest part of the requested period. Eleven machines recording every 30 seconds can produce 31,680 rows per day.

This hub-only fix streams the full selected period into per-machine chart aggregates. It avoids storing a decoded history slice and indexes cells by chart bucket to avoid repeatedly scanning the entire aggregate. The default query needs no global sort. Request cancellation and an eight-second read deadline return an error instead of a successful partial response.

Explicit max_records diagnostic requests remain supported. Raw-row lookahead distinguishes an exact fit from actual truncation even with corrupt payloads; deterministic ordering and a one-bin buffer withhold a bucket split by the cap. The reported boundary changes without shifting the original chart grid.

Power classifications, measured/modelled separation, current snapshots, coverage qualifications, and the withheld energy/cost UI are unchanged. No agent change, release, migration, or deployment.

Validation:
- Full hub suite: PASS (124.989s); go vet: PASS.
- Targeted race tests for the new scenarios: PASS (19.653s).
- Real SQLite/handler test: 63,360 rows, 22 machines, all 96 chart bins; matches complete reference aggregation with measured, legacy modelled, unknown-scope and gap inputs.
- Exact-fit, corrupt-row, partial-bin, unchanged-grid, and cancelled-request regression cases.
- Control: temporarily restoring LIMIT 20000 makes the full-day endpoint test fail with “streamed endpoint differs from complete reference aggregation”; restored code passes.
- Aggregation-only benchmark, Apple M4 Max, one iteration: 11 machines ~30ms / 1.2MB allocated; 100 machines ~290ms / 11.4MB allocated. These exclude SQLite/JSON costs and are not production latency measurements.
- gofmt and git diff --check: clean.

Claude and Kimi independently reviewed commit f5cbc3f and both approved with no blockers. Claude also approved the final defensive consumer-stop fix at 28bc330; focused fleet-power tests and vet passed on that commit. Unexpected early termination now preserves a cursor error or returns an explicit error, while deliberate capped reads remain successful. All seven CI checks passed on final head 28bc3309eb97f8c16d579a9ec047069fabe9adb6: Hub Tests, Hub Tests (race), Agent Tests, Agent Tests (windows), Dashboard Lint + Build, Shared Protocol Tests, and Compose smoke. No production verification or deployment performed.


bokiko:
@claude

